PR #1, opened 2017-01-25
DM-8798: Ingest HiTS calibration products and test on raw images

mrawls:
Note: some of these commits have different usernames because git knew me as "mrawls" on lsst-dev and "Meredith Rawls" on my laptop. I have fixed it so both places use my full name.

### DIFF
--- a/hits_ingest.py
+++ b/hits_ingest.py
@@ -7,6 +7,7 @@ from lsst.pipe.tasks.ingestCalibs import IngestCalibsConfig, IngestCalibsTask
 from lsst.pipe.tasks.ingestCalibs import IngestCalibsArgumentParser
 from lsst.pipe.tasks.processCcd import ProcessCcdTask
 from lsst.pipe.base import ArgumentParser
+from lsst.utils import getPackageDir
 from glob import glob
 import sys
 import sqlite3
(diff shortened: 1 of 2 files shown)


hits_ingest.py
```diff
@@ -1,60 +1,105 @@
...
-#calibrepodir = '../calibingested/' # on lsst-dev
-repodir = 'ingested/' # on laptop
-calibrepodir = 'calibingested/' # on laptop
-
-datadir = sys.argv[1] # or '/lsst7/mrawls/HiTS/Blind15A_38/' on lsst-dev or 'data/' on laptop
-datafiles = glob(datadir + '*.fits.fz')
-
-# first, need to make a text file that handles the mapper, per the obs_decam github README
-f = open(repodir+'_mapper', 'w')
+# edit values below as desired
```
ctslater: I think this block below here could basically be replaced with argparse and would be a bit more functional. Plus, you could make the `--help` text print the big message you have above, which would be very handy.

hits_ingest.py
```diff
@@ -1,60 +1,105 @@
...
+    datadir = sys.argv[1]  # '/lsst7/mrawls/HiTS/<dirname>' on lsst-dev
+                       # 'data/' or 'MasterCals/' on laptop
+                       # (if doIngestCalibs, datadir must be the dir of calibs)
+                       # can also use a regex, e.g., 'data/c4d_15*.fz'
+except:
+    datadir = None
+    print('WARNING: no data directory specified.')
+    print('(This is OK if you are just running ProcessCcd.)')
+else:    
+    if datadir[-1] == '/':
```
ctslater: Instead of relying on a `/`, which I would invariably forget had meaning, it might be nicer to use something like `os.path.isdir()`.

hits_ingest.py
```diff
@@ -1,60 +1,105 @@
...
+    else:
+        datafiles = glob(datadir)
+
+# make a text file that handles the mapper, per the obs_decam github README
+f = open(repo+'_mapper', 'w')
 print('lsst.obs.decam.DecamMapper', file=f)
 f.close()
 
-if doIngest == True:
+if doIngest:
```
mrawls: I agree; I plan to implement this going forward, but not for this ticket.

ctslater: Right, I forgot to mention that I'm totally on board with some of these being "out of scope" for this ticket. No worries

hits_ingest.py
```diff
@@ -64,25 +109,56 @@
...
+        print('Success!')
+        print('Calibrations corresponding to {0} are now ingested in {1}' \
+                                                    .format(repo, calibrepo))
+
+# the final step is to run processCcd to do ISR by combining the ingested
+# images and calibration products
+if doProcessCcd:
+    print('Running ProcessCcd...')
+    OBS_DECAM_DIR = os.getenv('OBS_DECAM_DIR')
+    os.chdir(calibrepo)
```
mrawls: As far as I can tell, this is a consequence of DM-5466, which I really do plan to tackle soon. I don't like having the chdir in here either so I will add a comment/note to revisit this once obs_decam is a little smarter about calib paths.

hits_ingest.py
mrawls: I'm not very familiar with argparse, but I will find some examples and keep this in mind going forward.

PR #2, opened 2017-05-05
DM-9654: Extend hits_ingest script to handle all steps from ingestion to diffim

mrawls:


hits_ingest.py
```diff
@@ -5,88 +5,102 @@
...
-    $ processCcd.py ../repo --id visit=visit ccdnum=ccdnum 
-            --output ../outputrepo -C $OBS_DECAM_DIR/config/processCcdCpIsr.py
-            --config calibrate.doAstrometry=False calibrate.doPhotoCal=False
-            --no-versions
+    $ processCcd.py repo --id visit=visit ccdnum=ccdnum
+            --output processedrepo --calib calibrepo
+            -C $OBS_DECAM_DIR/config/processCcdCpIsr.py
+            -C processccd_config.py
+            --config calibrate.doAstrometry=True calibrate.doPhotoCal=True
+    ** note that to run from bash, 'processccd_config.py' must exist
```
mrawls: Added explicit description of what should go here. However, I'm getting a 'no locations for get' error on the `pan-starrs refObjLoader.ref_dataset_name` now when I run this task, and would like your help resolving that. (It can't find gaia either; did these move?)

hits_ingest.py
```diff
@@ -5,88 +5,102 @@
...
-# edit values above as desired
-
-if ((doIngest and doIngestCalibs) or (doIngest and doProcessCcd) or 
-    (doIngestCalibs and doProcessCcd)):
-    raise RuntimeError('You can only have one task True at a time')
-elif not doIngest and not doIngestCalibs and not doProcessCcd:
-    raise RuntimeError('Nothing to do, every task is set to False')
-
-if len(sys.argv) < 2:
+# ~~  edit values below as desired  ~~ #
```
mrawls: I haven't implemented a yaml config file, but I did clean up the input significantly.

hits_ingest.py
```diff
@@ -141,24 +153,45 @@
...
+            '--calib', calibrepo,
             '-C', OBS_DECAM_DIR + '/config/processCcdCpIsr.py',
-            '--config', 'calibrate.doAstrometry=False',
-            'calibrate.doPhotoCal=False',
-            '--no-versions']
-    # This is SO MUCH EASIER than parsing the args in the above two cases!!
-    ProcessCcdTask.parseAndRun(args=args, doReturnResults=True)
-
-    
+            '--config', 'calibrate.doAstrometry=True',
```
mrawls: I believe they are True the default, but I'd like to argue to keep the explicit assignments because it differs from the example in the obs_decam README which has them both explicitly set to False. For difference imaging to work, they need to both be True.

hits_ingest.py
```diff
@@ -5,88 +5,102 @@
...
+from lsst.meas.algorithms import LoadIndexedReferenceObjectsTask
 from lsst.utils import getPackageDir
+from lsst.ip.diffim.getTemplate import GetCalexpAsTemplateTask
+from lsst.pipe.tasks.imageDifference import ImageDifferenceConfig, ImageDifferenceTask
 from glob import glob
-import sys
 import sqlite3
 import os
+import argparse
 '''
```
mrawls: I'm not sure what you're asking for here?

mrawls:
I refactored everything to have functions (yay!) and have tested ingest and ingestCalibs. I'm encountering an error (described above) when it's time to get references for astrometry, probably just because the reference catalogs moved and I'm not sure where it lives now. Once I fix this, I will run the processCcd and diffIm tasks and make sure they still work as intended.

I also removed the plotting program from this repo and renamed the main script (since it's not HiTS-specific) per your suggestions.

hits_ingest.py
ctslater: There are all these nice strings describing how to use the script, but right now the user has to open up the code to read them. If one could just call `python hits_ingest.py --help` and see some nice text, that would be helpful. Argparse supports this.

decam_process.py
```diff
@@ -0,0 +1,266 @@
...
+    if args.files is None:
+        datadir = None
+        datafiles = None
+    else:
+        datadir = args.files
+        if os.path.isdir(datadir):
+            datafiles = glob(os.path.join(datadir, '*.fz'))
+        else:
+            datafiles = glob(datadir)
+    main()
```
ctslater: I'm not particularly excited about a separate `main()` function (to me, `if __name__ == "__main__"` _is_ the main function), but that's just a personal opinion so I don't mind this choice. However, I am super grumpy about **implicitly** passing variables to functions. This obscures the behavior on both ends (what variables does this block expose to `main()`, and what variables in `main()` are intended to be inherited from here?). I'd strongly recommend either merging `main()` back here or calling out the variables that need to be passed as explicit function arguments.

mrawls: OK, that makes sense. I'd like to have the "main stuff" upfront, including calling the functions that follow, for clarity. To be less implicit but still support this structure, I changed the call to main at the end to include _just_ `main()`.

mrawls:
I have confirmed the script in its current form works for all four tasks (ingest and ingestCalibs run yesterday; processCcd and diffIm run today - hooray symlinking reference catalogs!). Unless you have more comments, I think I'm ready to call this done.

PR #3, opened 2017-07-21
DM-10977:  Share ImageSelector from Twinkles

ebellm:
Add helper code for making best-seeing coadds needed for `verify_ap` testing.


ups/decam_hits.table
```diff
@@ -0,0 +1,16 @@
...
+# rather than assuming it will be included recursively.
+setupRequired(obs_decam)
+
+# The following is boilerplate for all packages.
+# See Tech Note DMTN-001 for details on LSST_LIBRARY_PATH
+# Pure-Python packages only need the last line.
+envPrepend(LD_LIBRARY_PATH, ${PRODUCT_DIR}/lib)
+envPrepend(DYLD_LIBRARY_PATH, ${PRODUCT_DIR}/lib)
+envPrepend(LSST_LIBRARY_PATH, ${PRODUCT_DIR}/lib)
+envPrepend(PYTHONPATH, ${PRODUCT_DIR}/python)
```
kfindeisen: No `decam_hits.cfg` file.

kfindeisen: These `envPrepend` commands seem pointless -- there is no `lib` directory, and in a Python-only package there shouldn't be.

python/lsst/decam/hits/__init__.py
```diff
@@ -0,0 +1,7 @@
+from __future__ import absolute_import
+
+#
+# See COPYRIGHT file at the top of the source tree
+#
```
kfindeisen: IIRC we're not supposed to use this comment (RFC pending lawyer approval).

config/makeCoaddTempExp_goodSeeing.py
```diff
@@ -0,0 +1,22 @@
+from lsst.decam.hits import MaxPsfWcsSelectImagesTask
+
+config.bgSubtracted=True
```
kfindeisen: `config` is never defined; how does this not cause a `NameError`?

config/makeCoaddTempExp_goodSeeing.py
```diff
@@ -0,0 +1,22 @@
...
+
+# changing coaddName from deep requires modifying the obs_ package--wait 
+# until after upcoming RFD
+#config.coaddName='goodSeeing'
+
+config.select.retarget(MaxPsfWcsSelectImagesTask)
+config.doPsfMatch=True
+# !! This must be the same as the kernelSize in the processEimage psfMeasurement.
+config.modelPsf.size=25
```
kfindeisen: Can the matching constraint be enforced in code (e.g., `config.modelPsf.size=processEimage.config.kernelSize`)?

python/lsst/decam/hits/selectImages.py
```diff
@@ -0,0 +1,101 @@
...
+        We use "convexHull" instead of generating a SphericalConvexPolygon
+        directly because the standard for the inputs to SphericalConvexPolygon
+        are pretty high and we don't want to be responsible for reaching them.
+        If "convexHull" is found to be too slow, we can revise this.
+
+        @param dataRef: Data reference for coadd/tempExp (with tract, patch)
+        @param coordList: List of Coord specifying boundary of patch
+        @param makeDataRefList: Construct a list of data references?
+        @param selectDataList: List of SelectStruct, to consider for selection
+        """
```
kfindeisen: Please follow the documentation style guide (in particular, numpy style, and move implementation details to a `Notes` section or an ordinary comment).

python/lsst/decam/hits/selectImages.py
```diff
@@ -0,0 +1,101 @@
...
+        from lsst.geom import convexHull
```
kfindeisen: I think local imports are discouraged by PEP8?

python/lsst/decam/hits/selectImages.py
```diff
@@ -0,0 +1,101 @@
...
+            dataRef = data.dataRef
+            imageWcs = data.wcs
+            cal = dataRef.get('calexp', immediate=True)
+            psf_size = cal.getPsf().computeShape().getDeterminantRadius()
+            nx, ny = data.dims
+
+            imageBox = afwGeom.Box2D(afwGeom.Point2D(0, 0), afwGeom.Extent2D(nx, ny))
+            try:
+                imageCorners = [imageWcs.pixelToSky(pix) for pix in imageBox.getCorners()]
+            except (pexExceptions.DomainError, pexExceptions.RuntimeError) as e:
```
kfindeisen: Please try to make this more specific. You're essentially catching `pexExceptions.Exception` here; does `Wcs.pixelToSky` not document what it throws?

python/lsst/decam/hits/selectImages.py
```diff
@@ -0,0 +1,101 @@
...
+            if patchPoly.intersects(imagePoly):  # "intersects" also covers "contains" or "is contained by"
+                psf_sizes.append(psf_size)
+                dataRefList.append(dataRef)
+                exposureInfoList.append(BaseExposureInfo(dataRef.dataId, imageCorners))
+
+        filteredDataRefList = []
+        filteredExposureInfoList = []
+        for size_sigma, dataRef, expInfo in zip(psf_sizes, dataRefList, exposureInfoList):
+            # size_sigma is in sigma.  Convert to FWHM
+            size_fwhm = size_sigma*2.355
```
kfindeisen: Why use a truncated constant instead of the exact value (√(8 ln 2))?

python/lsst/decam/hits/selectImages.py
```diff
@@ -0,0 +1,101 @@
...
+            psf_size = cal.getPsf().computeShape().getDeterminantRadius()
+            nx, ny = data.dims
+
+            imageBox = afwGeom.Box2D(afwGeom.Point2D(0, 0), afwGeom.Extent2D(nx, ny))
+            try:
+                imageCorners = [imageWcs.pixelToSky(pix) for pix in imageBox.getCorners()]
+            except (pexExceptions.DomainError, pexExceptions.RuntimeError) as e:
+                # Protecting ourselves from awful Wcs solutions in input images
+                self.log.debug("WCS error in testing calexp %s (%s): deselecting", dataRef.dataId, e)
+                continue
```
kfindeisen: Are you sure it's safe to silently ignore corrupted data? I'm especially worried about what will happen if this Task returns empty lists.

ups/decam_hits.table
ebellm: These were fixed in other commits during the transition to ap_pipe.

python/lsst/decam/hits/__init__.py
ebellm: Also no longer relevant.

config/makeCoaddTempExp_goodSeeing.py
ebellm: I don't know, but I think it's handled by CommandLineTask?  See https://pipelines.lsst.io/v/DM-11597/modules/lsst.pipe.base/command-line-task-config-howto.html

python/lsst/decam/hits/selectImages.py
ebellm: This looks totally standard to me, and I don't find anything relevant in PEP8--can you point me to it?

python/lsst/decam/hits/selectImages.py
ebellm: Fixed.

python/lsst/decam/hits/selectImages.py
kfindeisen: I think I was thinking of the bit in https://www.python.org/dev/peps/pep-0008/#imports that says "Imports are always put at the top of the file", and the analogous text in the DM style guide. It's not explicitly forbidding local imports, but implies they should be kept together.

python/lsst/decam/hits/selectImages.py
kfindeisen: Oops, objection withdrawn. I thought it said `LogicError`, not `DomainError`. :worried:

python/lsst/decam/hits/selectImages.py
ebellm: Ah, I understand now.  Fixed.

config/makeCoaddTempExp_goodSeeing.py
ebellm: I actually think the right thing to do here is to remove this line entirely.  `processEimage` is from `obs_lsstSim`, and it doesn't appear to me that `makeCoaddTempExp` actually uses the `size` parameter anymore.  According to the docs, `defaultFwhm` is the parameter that is used, and it needs to be tuned to the individual dataset.  I had already been passing `defaultFwhm` from the command line.

python/lsst/decam/hits/selectImages.py
ebellm: Empirically, it just returns

```makeCoaddTempExp WARN: No exposures to coadd for patch DataId(initialdata={'filter': 'g', 'patch': '0,0', 'tract': 0}, tag=set([]))```

Which I think is fine.

kfindeisen: You mean, the rest of the pipeline just does nothing until it finishes successfully?

ebellm: `selectImages` is passing `dataRef`s up the `makeCoaddTempExp`.  If it returns empty lists because the data is bad, `makeCoaddTempExp` warns that there is no data for that patch and continues along.  (This can also happen if there is in fact no data there.)

kfindeisen: I would have thought that "no data" should be something alarming, but I think you're saying this is standard policy. Fine, objection withdrawn.

PR #4, opened 2017-07-28
DM-11021: Generalize prototype pipeline script

mrawls:


decam_process.py
```diff
@@ -15,161 +15,82 @@
...
+    processedrepo = os.path.join(args.output, 'processed')
+    diffimrepo = os.path.join(args.output, 'diffim')
+    types = ('*.fits', '*.fz')
+    datafiles = []
+    allcalibdatafiles = []
+    for files in types:
+        datafiles.extend(glob(os.path.join(args.dataset, 'raw', files)))
+        allcalibdatafiles.extend(glob(os.path.join(args.dataset, 'calib', files)))
+
+    # Ignore wtmaps and illumcors
```
kfindeisen: You might consider breaking up `main()` into several functions, e.g., one for argument parsing, one for finding the files, one for the "run everything" bit... Stuff could be divided past that; for example, the code used to create `datafiles` and `allcalibdatafiles` could be split into a single function that takes a directory and an extension.

decam_process.py
```diff
@@ -15,324 +15,378 @@
...
+    organized by date (bias and flat images only)
+    sqlite3 database registry of ingested calibration products (bias, flat,
+    and defect images) created in calibrepo
+
+    NOTE:
+    calib ingestion ingests *all* the calibs, not just the ones needed
+    for the specified visits. We may want to revisit this in the future.
     '''
+
+    def flatBiasIngest(repo, calibrepo, calibdatafiles):
```
djreiss: This function would be nice to have at the module level instead of as an internal function. Same for `defectIngest`.

decam_process.py
```diff
@@ -15,324 +15,378 @@
...
+            ingestTask.run(parsedCmd)
+        except sqlite3.IntegrityError as detail:
+            print('sqlite3.IntegrityError: ', detail)
+            print('(sqlite3 doesn\'t think all the calibration files are unique)')
+        else:
+            print('Success!')
+            print('Calibrations corresponding to {0} are now ingested in {1}'.format(repo, calibrepo))
+        return
+
+    def defectIngest(repo, calibrepo, defectfiles):
```
djreiss: I don't know a whole lot about the ingestion process. Which got me thinking -- there should be some unit tests for all of this code.

decam_process.py
```diff
@@ -15,324 +15,378 @@
...
+        Ingest DECam defect images
+    
+        Parameters
+        ----------
+        repo: `str`
+            The output repository location on disk where ingested raw images live.
+        calibrepo: `str`
+            The output repository location on disk where ingested calibration images live.
+        defectfiles: `list`
+            A list of the filenames of each defect image file.
```
djreiss: Confused - why a list of filenames if we supply a tarball containing all of them?

decam_process.py
```diff
@@ -15,324 +15,378 @@
...
+        try:
+            os.mkdir('defects')
+        except OSError:
+            # defects directory already exists
+            print('Defects were previously ingested, skipping...')
+        else:
+            print('Ingesting defects...')
+            defectargs = ['../../' + repo, '--calib', '.', '--calibType', 'defect',
+                          '--mode', 'skip', '--validity', '999']
+            defecttarball = defectfiles[0] + '.tar.gz'
```
djreiss: Hate to be nit-picky but I'd prefer to see this variable named `defectTarball` which makes it a bit easier to read. Same with other variable names.

decam_process.py
```diff
@@ -15,324 +15,378 @@
...
+            tarfile.open(os.path.join('../../', defecttarball), 'r').extractall('defects')
+            defectfiles = glob(os.path.join('defects', os.path.basename(defectfiles[0]), '*.fits'))
+            defectargs.extend(defectfiles)
+            defectargumentParser = IngestCalibsArgumentParser(name='ingestCalibs')
```
djreiss: `defectArgumentParser`, etc...

decam_process.py
```diff
@@ -15,324 +15,378 @@
...
-    except sqlite3.IntegrityError as detail:
-        print('~~~ !!! ~~~')
-        print('sqlite3.IntegrityError: ', detail)
-        print('(sqlite3 doesn\'t think all the calibration files are unique)')
-        print('If this isn\'t your first time ingesting these calibration')
-        print('  files, move or delete the existing database and try again.')
-        print('If this is your first time ingesting these calibration files,')
-        print('  make sure you only use image or wtmap MasterCals, not both.')
-        print('~~~ !!! ~~~')
+        flatBiasIngest(repo, calibrepo, calibdatafiles)
```
djreiss: Maybe not for this ticket but I think all of this code would benefit from going into a class, instead of having passing around all the different variables from one function to the next.

decam_process.py
```diff
@@ -15,324 +15,378 @@
...
-        for refObjLoader in (config.calibrate.astromRefObjLoader,
-                            config.calibrate.photoRefObjLoader,
-                            config.charImage.refObjLoader,
-                            ):
-        refObjLoader.retarget(LoadIndexedReferenceObjectsTask)
-        refObjLoader.ref_dataset_name = 'pan-starrs'
-        refObjLoader.filterMap = {"g": "g",
-                                  "r": "r",
-                                  "VR": "g"}
+    ** to run from bash, 'processccd_config.py' must exist and contain
```
djreiss: What is the equivalent of using the `processccd_config.py` file for this script? Is it hardcoded somewhere, or not used?

decam_process.py
```diff
@@ -15,324 +15,378 @@
...
-                                  "VR": "g"}
-    args = [repo, '--id', 'visit=' + visit, 'ccdnum=' + ccdnum,
-            '--output', processedrepo,
-            '--calib', calibrepo,
-            '-C', OBS_DECAM_DIR + '/config/processCcdCpIsr.py',
-            '--config', 'calibrate.doAstrometry=True',
-            'calibrate.doPhotoCal=True']
-    ProcessCcdTask.parseAndRun(args=args, config=config)
+    if os.path.isdir(os.path.join(processedrepo, '0'+visit)):
+        print('ProcessCcd has already been run for visit {0}, skipping...'.format(visit))
```
djreiss: Again, maybe add a `return` statement here and move everything else out of the `else` block.

decam_process.py
djreiss: Nevermind, I see it below. Might be good to enumerate the default values that are set in the config (e.g. using gaia for astrometric reference, etc.)

decam_process.py
```diff
@@ -15,324 +15,378 @@
...
 def doDiffIm(processedrepo, sciencevisit, ccdnum, templatevisit, diffimrepo):
     '''
     Do difference imaging with a visit as a template and one or more as science
-
-    USAGE:
-    $ python hits_ingest.py diffIm
+    
+    Parameters
+    ----------
+    processedrepo: `str`
```
djreiss: I first read this as `proccess drepo`. I like camel-case or underscore (I don't care which but I think the style guide has us moving toward using underscore, e.g. `processed_repo`.

decam_process.py
```diff
@@ -15,324 +15,378 @@
...
+    $ python decam_process.py -d dataset_root -o output_location -i "visit=12345, ccd=5"
                                      '''))
-    parser.add_argument('task', choices=['ingest', 'ingestCalibs', 'processCcd', 'diffIm'],
-                        help=textwrap.dedent('''
-    Which of four tasks you would like to run.
-
-    ingest - Ingest raw DECam images into a repository with a
-    corresponding registry.
-    USAGE:
-    $ python hits_ingest.py ingest -f path/to/rawimages/
```
mrawls: Essentially yes; a call to run the pipeline script via `main()` will run all of the steps rather than just one at a time. This doesn't matter for `ap_verify` (or any other external code using the pipeline) because it can call whichever function it wants whenever it wants. Is that OK?

decam_process.py
```diff
@@ -15,324 +15,378 @@
...
-    diffimrepo/deepDiff/v+sciencevisit populated with difference images
-    and catalogs of detected sources (diaSrc, diffexp, and metadata files)
-                        '''))
-    parser.add_argument('-f', '--files',
-                        help='Input files or directory for ingest or ingestCalibs.')
+    parser.add_argument('-d', '--dataset',
+                        help="Location on disk of dataset_root, which contains subdirectories of raw data, calibs, etc.")
+    parser.add_argument('-o', '--output',
+                        help="Location on disk where output repos will live.")
+    # TODO: implement this argument
```
mrawls: It is part of DM-11422 because using coadds as templates will require `--id`-like metadata to be handled differently than when a visit is a template. I'll note this explicitly here.

decam_process.py
mrawls: Fair enough! Mucking with `main()` is a bit lower on my priority list since `ap_verify` doesn't care about it, but you're both correct that this function is rather lengthy with multiple functionalities. I'll see what I can do.

decam_process.py
mrawls: I was wondering when you'd say that. I realized the same thing and have started writing some!

decam_process.py
kfindeisen: For future reference: https://developer.lsst.io/coding/python_style_guide.html#to-do-comments-should-include-a-jira-issue-key :)

decam_process.py
kfindeisen: Should these names be camel-case or underscore-separated? PEP8 is kind of vague about local variables.

decam_process.py
kfindeisen: Can you be more specific about the benefits? You made a similar argument for `ap_verify` and I'm still not sure it was the right thing to do; in particular, explicit function arguments make it much easier to see (and control) each task's data dependencies.

mrawls: Happy to discuss with both of you further, but "not for this ticket" indeed!

decam_process.py
```diff
@@ -15,324 +15,378 @@
...
-    elif args.task == 'diffIm':
-        doDiffIm(processedrepo, sciencevisit, ccdnum, templatevisit, diffimrepo)
+
+    if not os.path.isdir(args.output):
+        os.mkdir(args.output)
+    repo = os.path.join(args.output, 'ingested')
+    calibrepo = os.path.join(args.output, 'calibingested')
+    processedrepo = os.path.join(args.output, 'processed')
+    diffimrepo = os.path.join(args.output, 'diffim')
+    types = ('*.fits', '*.fz')
```
mrawls: I've never seen a `fits.gz` file, is that a common thing and are they treated just like fits files throughout the stack (much like it seems `fits.fz` files are)?

djreiss: I dunno. Never mind ;)

decam_process.py
mrawls: I agree this isn't very elegant, but I did this so the pipeline expects the list of defect files in the same way the regular calibs are supplied as lists of files. Would it be clearer if `doIngestCalibs` just took the name of the defect tarball?

kfindeisen: I think your way is better -- not only is it easier to understand, but the code's not dependent on defects being provided as exactly one tarball.

djreiss: Sounds good to me.

mrawls:
Thanks for the thorough review! I think the only outstanding items are (1) adding tests and (2) possibly changing how defect file/tarball info is passed into doIngestCalibs. I would like to close this ticket so I can rename/reorganize the repo to finally be `ap_pipe` before making many more changes to the code. I propose handling these outstanding items as part of DM-11390 if that is OK with you.

decam_process.py
```diff
@@ -207,158 +246,307 @@ def doIngest(repo, datafiles):
     config = IngestConfig()
     config.parse.retarget(DecamParseTask)
     # create an instance of the decam ingest task
-    ingestTask = ingest.DecamIngestTask(config=config)
+    IngestTask = ingest.DecamIngestTask(config=config)
```
djreiss: Is there a reason you changed this to capitalized? Capitalized names are for classes, whereas this is a class instance?

decam_process.py
```diff
@@ -207,158 +246,307 @@ def doIngest(repo, datafiles):
...
-    args = [repo, '--calib', calibrepo, '--mode', 'link', '--validity', '999']
-    args.extend(datafiles)
+    print('Ingesting flats and biases...')
+    args = [repo, '--calib', calib_repo, '--mode', 'link', '--validity', '999']
+    args.extend(calibdatafiles)
     argumentParser = IngestCalibsArgumentParser(name='ingestCalibs')
     config = IngestCalibsConfig()
     config.parse.retarget(ingestCalibs.DecamCalibsParseTask)
-    ingestTask = IngestCalibsTask(config=config, name='ingestCalibs')
+    CalibIngestTask = IngestCalibsTask(config=config, name='ingestCalibs')
```
djreiss: Same comment as above.

decam_process.py
```diff
@@ -207,158 +246,307 @@ def doIngest(repo, datafiles):
...
+    This function assumes very particular things about defect ingestion:
+    - They must live in a .tar.gz file in the same location on disk as the other calibs
+    - They will be ingested using ingestCalibs.py run from the calib_repo directory
+    - They will be manually uncompressed and saved in calib_repo/defects/<tarballname>/.
+    - They will be added to the calib registry, but not linked like the flats and biases
+    '''
+    os.chdir(calib_repo)
+    try:
+        os.mkdir('defects')
+    except OSError:
```
djreiss: I like the change over to try/except/else!

kfindeisen: This should be `IOError` -- `OSError` is not used for file I/O problems in Python 2, but `IOError` is an alias for `OSError` in Python 3.

kfindeisen: Never mind, it's always `OSError`. :astonished: 

You'll probably brand this as "not this ticket", but I don't think it's safe to assume a pre-existing directory is the only reason `os.mkdir` raises -- it could also be lack of permissions, problems with the filesystem itself, etc...

decam_process.py
mrawls: Probably because the story of my life is being confused about what is a class and what is an instance... I'll change it back :)

decam_process.py
mrawls: Fair point @kfindeisen, I will make the statement that prints with that error mention the fact that the exception is _likely_ (but not necessarily) happening because the defect directory already exists.

PR #5, opened 2017-08-11
DM-11325: Make prototype pipeline eups-compatible

mrawls:


.gitignore
```diff
@@ -1,4 +1,19 @@
-ingested/*
-calibingested/*
-data/*
-MasterCals/*
```
kfindeisen: Vestigial from before we had `ap_verify_hits2015`, I assume?

doc/conf.py
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env python
+
+from documenteer.sphinxconfig.stackconf import build_package_configs
+
+import lsst.afw
+
+_g = globals()
+_g.update(build_package_configs(
+    project_name="afw",
+    copyright="2017 Association of Univerities for "
```
kfindeisen: Oops! There's another `afw` reference below.

ups/ap_pipe.table
```diff
@@ -0,0 +1,23 @@
...
+# Any other package whose interface is used should be listed explicitly
+# rather than assuming it will be included recursively.
+setupRequired(boost >= 1.47.0)
+setupRequired(utils >= 4.6.0.0)
+setupRequired(pex_exceptions >= 4.6.0.0)
+setupRequired(ndarray)
+
+setupRequired(obs_decam)
+setupOptional(ap_verify_hits2015)
+setupOptional(ap_verify)
```
kfindeisen: Definitely shouldn't have `ap_verify` here, because that gives a circular dependency. I don't think `ap_verify_hits2015` is appropriate either.

ups/ap_pipe.cfg
```diff
@@ -0,0 +1,26 @@
...
+
+# Dependencies that provide header files and or libraries should be included here.
+# Pure-Python dependencies do not need to be included.
+# Packages that use pybind11 or boost_tests should declare them as build dependencies.
+# Otherwise, the rules for which packages to list here are the same as those for
+# table files.
+dependencies = {
+    "required": ["utils", "ndarray"],
+    "buildRequired": ["boost_test", "pybind11", "ndarray", "pex_exceptions", "obs_decam"],
+    "optional": ["ap_verify_hits2015", "ap_verify"],
```
kfindeisen: I don't think `ap_pipe.cfg` needs to contain many packages, particularly not Python-only ones, because it's used by Scons to set up paths and the like. But I'll admit I don't know for sure.

doc/conf.py
kfindeisen: I just did a dumb substitution of `afw` -> `ap_verify`/`ap.verify` on my end and it doesn't seem to have broken documentation building...

ups/ap_pipe.cfg
mrawls: I saw yours was empty for `dependencies`; I went with the one in the template repo and just added `obs_decam`. But I suspect you're right about not needing to have anything in here for python-only dependencies, so I'll make it empty instead.

doc/conf.py
mrawls: Good catch 👍

PR #6, opened 2017-09-21
DM-11390: Plug prototype pipeline script into verify_ap framework

mrawls:


python/lsst/ap/pipe/ap_pipe.py
```diff
@@ -45,160 +57,207 @@
...
-    from the command line, on a dataset intended for ap_verify. It is useful
-    for testing or standalone image processing independently from verification.
+    Returns
+    -------
+    repos_and_files: `dict`
+        Includes the names of new repos that will be written to disk
+        following ingestion, calib ingestion, processing, and difference imaging
+        (repo, calib_repo, processed_repo, diffim_repo)
+        Includes the files in dataset_root for raw images, flats and biases,
+        and defects (datafiles, calibdatafiles, defectfiles)
```
kfindeisen: Might want to have these names in quotes just to make it obvious that they're string keys.

python/lsst/ap/pipe/ap_pipe.py
```diff
@@ -45,160 +57,207 @@
...
+    idlist: `list` containing two `str`
+        Data ID and template info needed for processing and difference imaging
+        [dataId, template]
+        TODO: allow 'template' to be either a visit ID or a repo name (DM-11422)
+    refcats: `str`
+        Path on disk of the reference catalogs
```
kfindeisen: Why are `idlist` and `refcats` separate from the other results?

python/lsst/ap/pipe/ap_pipe.py
```diff
@@ -221,16 +280,17 @@ def doIngest(repo, ref_cats, datafiles):
     ASTROM_REFCAT_DIR = 'ref_cats/gaia'
     PHOTOM_REFCAT_DIR = 'ref_cats/pan-starrs'
 
+    lsst.log.configure()
+    log = lsst.log.Log.getLogger('ap.pipe.doIngest')
     if os.path.exists(os.path.join(repo, 'registry.sqlite3')):
-        print('Raw images were previously ingested, skipping...')
+        log.warn('Raw images were previously ingested, skipping...')
```
kfindeisen: Loggers FTW :+1:

python/lsst/ap/pipe/ap_pipe.py
```diff
@@ -279,7 +339,9 @@ def flatBiasIngest(repo, calib_repo, calibdatafiles):
     BASH EQUIVALENT:
     $ ingestCalibs.py repo --calib calib_repo --mode=link --validity 999 calibdatafiles
     '''
-    print('Ingesting flats and biases...')
+    lsst.log.configure()
```
kfindeisen: Shouldn't this call live in your `main` function? Why should `flatBiasIngest` be responsible for setting up logging?

python/lsst/ap/pipe/ap_pipe.py
```diff
@@ -290,12 +352,12 @@ def flatBiasIngest(repo, calib_repo, calibdatafiles):
     try:
         calibIngestTask.run(parsedCmd)
     except sqlite3.IntegrityError as detail:
-        print('sqlite3.IntegrityError: ', detail)
-        print('(sqlite3 doesn\'t think all the calibration files are unique)')
+        log.error('sqlite3.IntegrityError: ', detail)
+        log.error('(sqlite3 doesn\'t think all the calibration files are unique)')
         flatBias_metadata = None
```
kfindeisen: Umm, `IntegrityError` sounds serious. Are you sure you can ignore it?

(Also, I think your first error message will redundantly give the name of the exception twice.)

python/lsst/ap/pipe/ap_pipe.py
```diff
@@ -331,18 +393,20 @@ def defectIngest(repo, calib_repo, defectfiles):
     - They will be manually uncompressed and saved in calib_repo/defects/<tarballname>/.
     - They will be added to the calib registry, but not linked like the flats and biases
     '''
+    lsst.log.configure()
```
kfindeisen: You should only [need to call this once](https://lsst-web.ncsa.illinois.edu/doxygen/x_masterDoxyDoc/log.html#initialization).

python/lsst/ap/pipe/ap_pipe.py
```diff
@@ -331,18 +393,20 @@ def defectIngest(repo, calib_repo, defectfiles):
...
     try:
         os.mkdir('defects')
     except OSError:
         # most likely the defects directory already exists
         if os.path.isdir('defects'):
-            print('Defects were previously ingested, skipping...')
+            log.warn('Defects were previously ingested, skipping...')
         else:
-            print('Defect ingestion failed because \'defects\' dir could not be created')
+            log.error('Defect ingestion failed because \'defects\' dir could not be created')
```
kfindeisen: Again, is it safe to continue like this?

python/lsst/ap/pipe/ap_pipe.py
```diff
@@ -394,23 +457,32 @@ def doIngestCalibs(repo, calib_repo, calibdatafiles, defectfiles):
 
     NOTE:
     calib ingestion ingests *all* the calibs, not just the ones needed
-    for the specified visits. We may want to revisit this in the future.
+    for certain visits. We may want to ...revisit... this in the future.
```
kfindeisen: I'm ...confused...

python/lsst/ap/pipe/ap_pipe.py
```diff
@@ -394,23 +457,32 @@ def doIngestCalibs(repo, calib_repo, calibdatafiles, defectfiles):
...
     '''
+    lsst.log.configure()
```
kfindeisen: Redundant initialization

python/lsst/ap/pipe/ap_pipe.py
```diff
@@ -422,42 +494,54 @@ def doProcessCcd(repo, calib_repo, processed_repo, visit, ccdnum):
...
-       already contain the ref_cats (this is done during doIngest).
+       already contain the refcats (this is done during doIngest).
 
     RESULT:
     processed_repo/visit populated with subdirectories containing the
     usual post-ISR data (bkgd, calexp, icExp, icSrc, postISR).
     By default, the configuration for astrometric reference catalogs uses Gaia
     and the configuration for photometry reference catalogs uses Pan-STARRS.
     '''
+    lsst.log.configure()
```
kfindeisen: Redundant initialization

python/lsst/ap/pipe/ap_pipe.py
```diff
@@ -422,42 +494,54 @@ def doProcessCcd(repo, calib_repo, processed_repo, visit, ccdnum):
...
+    log = lsst.log.Log.getLogger('ap.pipe.doProcessCcd')
+    dataId_items = re.split('[ +=]', dataId)
+    dataId_dict = dict(zip(dataId_items[::2], dataId_items[1::2]))
```
kfindeisen: I'm not sure I understand this. Is this basically pairing up the even- and odd-indexed elements of `dataId_items`?

kfindeisen: Also, what happens if `dataId` doesn't follow the right format?

python/lsst/ap/pipe/ap_pipe.py
```diff
@@ -422,42 +494,54 @@ def doProcessCcd(repo, calib_repo, processed_repo, visit, ccdnum):
...
         os.mkdir(processed_repo)
-    print('Running ProcessCcd...')
+    log.info('Running ProcessCcd...')
     OBS_DECAM_DIR = getPackageDir('obs_decam')
+    calib_repo = os.path.abspath(calib_repo)
+    butler = dafPersist.Butler(inputs={'root': repo, 'mapperArgs': {'calibRoot': calib_repo}},
+                               outputs=processed_repo)
     config = ProcessCcdConfig()
+    config.load(OBS_DECAM_DIR + '/config/processCcd.py')
+    config.load(OBS_DECAM_DIR + '/config/processCcdCpIsr.py')
```
kfindeisen: Prefer `os.path.join`.

python/lsst/ap/pipe/ap_pipe.py
```diff
@@ -527,27 +604,73 @@ def doDiffIm(processed_repo, sciencevisit, ccdnum, templatevisit, diffim_repo):
         config.doDecorrelation = True
 
     TODO: use coadds as templates by default, not another visit (DM-11422).
+    Some of the comments in this function are placeholders for DM-11422 work.
```
kfindeisen: This line should probably not be in the documentation, since it's a statement about the source code rather than the behavior.

python/lsst/ap/pipe/ap_pipe.py
```diff
@@ -527,27 +604,73 @@ def doDiffIm(processed_repo, sciencevisit, ccdnum, templatevisit, diffim_repo):
...
 
     RESULT:
-    diffim_repo/deepDiff/v+sciencevisit populated with difference images
+    diffim_repo/deepDiff/v+visit populated with difference images
     and catalogs of detected sources (diaSrc, diffexp, and metadata files)
     '''
-    if os.path.exists(os.path.join(diffim_repo, 'deepDiff', 'v' + sciencevisit)):
-        print('DiffIm has already been run for visit {0}, skipping...'.format(sciencevisit))
+    lsst.log.configure()
```
kfindeisen: Redundant initialization

python/lsst/ap/pipe/ap_pipe.py
```diff
@@ -527,27 +604,73 @@ def doDiffIm(processed_repo, sciencevisit, ccdnum, templatevisit, diffim_repo):
...
+    log = lsst.log.Log.getLogger('ap.pipe.doDiffIm')
+    dataId_items = re.split('[ +=]', dataId)
+    dataId_dict = dict(zip(dataId_items[::2], dataId_items[1::2]))
+    if 'visit' not in dataId_dict.keys():
+        raise RuntimeError('The dataId string is missing \'visit\'')
+    else:  # save the visit number from the dataId
+        visit = dataId_dict['visit']
```
kfindeisen: Duplicate code with `doProcessCcd`, consider factoring.

python/lsst/ap/pipe/ap_pipe.py
```diff
@@ -527,27 +604,73 @@ def doDiffIm(processed_repo, sciencevisit, ccdnum, templatevisit, diffim_repo):
...
+    log.info('Running ImageDifference...')
     config = ImageDifferenceConfig()
-    config.getTemplate.retarget(GetCalexpAsTemplateTask)
+    config.getTemplate.retarget(GetCalexpAsTemplateTask)  # visit template config
+    # config.doSelectSources = False  # coadd template config
     config.detection.thresholdValue = 5.0
     config.doDecorrelation = True
-    args = [processed_repo, '--id', 'visit=' + sciencevisit, 'ccdnum=' + ccdnum,
-            '--templateId', 'visit=' + templatevisit, '--output', diffim_repo]
+    dataId = dataId.split(' ')
```
kfindeisen: Why are you splitting on three characters above but only on spaces here?

python/lsst/ap/pipe/ap_pipe.py
```diff
@@ -45,160 +57,211 @@
 from lsst.utils import getPackageDir
 from lsst.ip.diffim.getTemplate import GetCalexpAsTemplateTask
 from lsst.pipe.tasks.imageDifference import ImageDifferenceConfig, ImageDifferenceTask
+import lsst.daf.persistence as dafPersist
 
+# IN PROGRESS: figure out which of these ALL-CAPS VARIABLES are already known by
+# ap_verify and which need to be explicitly provided in a function here in ap_pipe
```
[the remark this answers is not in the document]
mrawls: Whoops, this comment should have been removed once this situation was solved.

python/lsst/ap/pipe/ap_pipe.py
```diff
@@ -45,160 +57,207 @@
...
+    -------
+    defectfiles: `list`
+        A list of the filenames of each defect image file.
+        The first element in this list will be the name of a .tar.gz file
+        which contains all the compressed defect images.
+    '''
     # Retrieve defect filenames from tarball
-    defectloc = os.path.join(args.dataset, DEFECT_DIR)
-    defect_tarfile_path = glob(os.path.join(defectloc, DEFECT_TARBALL))[0]
+    defect_tarfile_path = glob(os.path.join(defect_location, DEFECT_TARBALL))[0]
```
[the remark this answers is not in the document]
mrawls: This is an idiosyncrasy of tarballs - using glob lists all the files in the tarball, and the zeroth entry is the name of the tarball itself.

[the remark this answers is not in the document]
kfindeisen: I meant why can't you just have `os.path.join(defect_location, DEFECT_TARBALL)`, without the call to `glob`?

python/lsst/ap/pipe/ap_pipe.py
mrawls: Primarily because I think about them as different entities. I'll go ahead and add `refcats` to `repos_and_files`, but I'm not going to touch `dataId` and `template` right now because so much around their handling is likely to change in DM-11422.

python/lsst/ap/pipe/ap_pipe.py
mrawls: I presently have `lsst.log.configure()` in each of the main `ap_pipe` functions (seven total instances, in fact). I assumed that each function needed to have its own logging setup, but that might not be right.

kfindeisen: I'm pretty sure you're only supposed to do it once per program; the LSST documentation isn't particularly clear, but most loggers work that way.

python/lsst/ap/pipe/ap_pipe.py
mrawls: This change isn't particularly meaningful, it's just me being clever with the word visit :)
The gist of the comment is that the ingestion steps will ingest everything in the provided raw (and calib) dir, not just a subset of images as might be desired. I'm not sure how easy (or desirable) it would be to search for visit info in, e.g., FITS headers, and only ingest those images (and their associated calibs).

python/lsst/ap/pipe/ap_pipe.py
mrawls: OK nevermind :)

python/lsst/ap/pipe/ap_pipe.py
mrawls: Yep. If `dataId` has an odd number of elements, it will just omit the last one and carry on. This seems as reasonable as anything.

python/lsst/ap/pipe/ap_pipe.py
mrawls: Unfortunately, `dataId` handling has to be different here because difference imaging uses `parseAndRun` while processing uses `run`. In the future, both Tasks may use `run`.

python/lsst/ap/pipe/ap_pipe.py
mrawls: Point taken, but the duplication is minimal and `dataId` handling is less than ideal all around. Leaving it as-is for now.

python/lsst/ap/pipe/ap_pipe.py
kfindeisen: I'd prefer some kind of parse error, but as discussed in person parsing dataIds is not something we should have to do ourselves anyway. Maybe someday we can delegate it all to the Task framework...

python/lsst/ap/pipe/ap_pipe.py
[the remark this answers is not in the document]
mrawls: Hey what do you know, that works! 🥇

PR #13, opened 2018-02-06
DM-12938: ap_pipe crashes if --output is absolute path

kfindeisen:
The previous code required that ap_pipe's output repo be a relative path to an immediate subdirectory of the user's working directory, which caused crashes when an unaware user provided a different path.

### DIFF
--- a/python/lsst/ap/pipe/ap_pipe.py
+++ b/python/lsst/ap/pipe/ap_pipe.py
@@ -440,6 +440,10 @@ def defectIngest(repo, calib_repo, defectfiles):
     - They will be added to the calib registry, but not linked like the flats and biases
     '''
     log = lsst.log.Log.getLogger('ap.pipe.defectIngest')
+    absRepo = os.path.abspath(repo)
+    defect_tarball = os.path.abspath(defectfiles[0] + '.tar.gz')
+    startDir = os.path.abspath(os.getcwd())
+    # CameraMapper does not accept absolute paths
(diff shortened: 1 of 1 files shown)


python/lsst/ap/pipe/ap_pipe.py
```diff
@@ -440,6 +440,10 @@ def defectIngest(repo, calib_repo, defectfiles):
     - They will be added to the calib registry, but not linked like the flats and biases
     '''
     log = lsst.log.Log.getLogger('ap.pipe.defectIngest')
+    absRepo = os.path.abspath(repo)
+    defect_tarball = os.path.abspath(defectfiles[0] + '.tar.gz')
+    oldDir = os.path.abspath(os.getcwd())
```
morriscb: Why the name oldDir? This makes it sound like it will be removed later. Maybe add a comment here explaining this directory and it's relation to the absRepo path.

morriscb:
Looks good overall. However, a short explanation of what oldDir is an why it's there may be in order. Esspecially the plan is at some point to remove it.

python/lsst/ap/pipe/ap_pipe.py
kfindeisen: "Old" as in "before mucking around with `chdir`." Would "startDir" or "baseDir" be better?

morriscb: Yes, sounds good.

PR #14, opened 2018-02-08
DM-13451: Make ap_verify responsible for ingestion

kfindeisen:
This PR removes the ingestion code from `ap_pipe` (except for template "ingestion", which is kept as a workaround for DM-11865) and modifies the command-line interface of `ap_pipe` to accept a repository in the format matching the previous code's output. The default behavior when no template is provided has changed from assuming a specific visit ID to assuming templates are provided by the input repository.

Due to API changes, must be merged simultaneously with lsst-dm/ap_verify#20.


python/lsst/ap/pipe/ap_pipe.py
```diff
@@ -196,373 +99,66 @@ def parsePipelineArgs():
     # Parse command line arguments with argparse
     parser = argparse.ArgumentParser(formatter_class=argparse.RawTextHelpFormatter,
                                      description=textwrap.dedent('''
-    Process raw decam images with MasterCals from ingestion --> difference imaging
+    Process raw decam images with MasterCals from basic reduction --> difference imaging
```
mrawls: Keep this consistent with what the very first docstring says: "from basic processing --> source association"

PR #15, opened 2018-02-13
DM-13535: Accept idiomatic input repositories

kfindeisen:
The new code allows `ap_pipe` to accept 1-2 standard Butler repositories as input (so long as they are DECam data). Runs on repositories created by `ap_verify` can still be made by passing input = "\<repo>/ingested" and --calib = "\<repo>/calibingested".

### DIFF
--- a/python/lsst/ap/pipe/ap_pipe.py
+++ b/python/lsst/ap/pipe/ap_pipe.py
@@ -39,7 +39,6 @@ __all__ = ['get_output_repo',
 
 import os
 import argparse
-import textwrap
 import re
 
 import lsst.log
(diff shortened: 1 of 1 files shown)


python/lsst/ap/pipe/ap_pipe.py
```diff
@@ -105,8 +105,9 @@ def parsePipelineArgs():
     $ python ap_pipe.py input_location -t template_location -o output_location -i "visit=12345 ccdnum=5"
```
mrawls: Please update this usage example so it is consistent with the accepted arguments; we must have missed this when template handling was changed.

mrawls: Or maybe it is consistent, it's just difficult to show all the possible combinations of use cases; my mistake.

kfindeisen:
I already did those tests 🙂 . In addition to the examples you mentioned, I checked for regressions of [DM-12938](https://jira.lsstcorp.org/browse/DM-12938).

python/lsst/ap/pipe/ap_pipe.py
kfindeisen: You mean adding `--calib`, I assume? I can do that.

On the other hand, would you be ok with removing this line from the description? The output looks kind of weird:

    $ python bin/ap_pipe.py -h
    usage: ap_pipe.py [-h] [-o OUTPUT] [-i DATAID] [--no-skip]
                      [--templateId TEMPLATEID | -t TEMPLATEREPO]
                      input

    Process raw decam images with MasterCals from basic processing --> source association

    USAGE:
    $ python ap_pipe.py input_location -t template_location -o output_location -i "visit=12345 ccdnum=5"

mrawls: Goes to show how often I try running things with the `-h` flag! Sure, it's a bit redundant with the actual usage info that prints out 👍

PR #19, opened 2018-05-08
DM-14359: Fix data ID handling in ap_*

kfindeisen:
This PR fixes several bugs in how datarefs were passed to `ApPipeTask`. It also updates `ap_pipe` to current style standards and removes use of `future`.

### DIFF
--- a/python/lsst/ap/pipe/ap_pipe.py
+++ b/python/lsst/ap/pipe/ap_pipe.py
@@ -190,6 +190,7 @@ class ApPipeTask(pipeBase.CmdLineTask):
         # TODO: treat as independent jobs (may need SuperTask framework?)
         if templateIds is not None:
             for templateId in templateIds:
+                # templateId is typically visit-only; consider only the same raft/CCD/etc. as rawRef
                 rawTemplateRef = _siblingRef(rawRef, "raw", templateId)
                 calexpTemplateRef = _siblingRef(calexpRef, "calexp", templateId)
                 if "ccdProcessor" not in reuse or not calexpTemplateRef.datasetExists("calexp", write=True):
@@ -329,7 +330,11 @@ def _setupDatabase(configurable):
 
 
 def _siblingRef(original, datasetType, dataId):
-    """Construct a new dataRef from an old one.
+    """Construct a new dataRef using an existing dataRef as a template.
+
+    The typical application is to construct a data ID that differs from an
+    existing ID in one or two keys, but is more specific than expanding a
+    partial data ID would be.
 
     Parameters
     ----------
--- a/python/lsst/ap/pipe/ap_pipe.py
+++ b/python/lsst/ap/pipe/ap_pipe.py
@@ -159,7 +159,7 @@ class ApPipeTask(pipeBase.CmdLineTask):
         return configClass(**contents)
 
     @pipeBase.timeMethod
-    def run(self, rawRef, templateIds=None, reuse=[]):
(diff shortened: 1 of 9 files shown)


python/lsst/ap/pipe/ap_pipe.py
```diff
@@ -186,6 +182,10 @@ def run(self, rawRef, calexpRef, templateIds=None, reuse=[]):
             - differencer : output of `config.differencer.run` (`lsst.pipe.base.Struct` or `None`).
             - associator : output of `config.associator.run` (`lsst.pipe.base.Struct` or `None`).
         """
+        if reuse is None:
+            reuse = []
```
parejkoj: Good catch and fix. Though you should document the None choice in the docstring.

python/lsst/ap/pipe/ap_pipe.py
```diff
@@ -334,13 +334,18 @@ def _siblingRef(original, datasetType, dataId):
...
         The desired type of the new dataRef. Must be compatible
         with ``original``.
     dataId : `dict` from `str` to any
         A possibly partial data ID for the new dataRef. Any properties left
         unspecified shall be copied from ``original``.
     """
+    # Work around mismatched HDU lists for raw and processed data
+    cleanId = original.dataId.copy()
+    if 'hdu' in cleanId:
+        del cleanId['hdu']
```
parejkoj: This method shouldn't be necessary: `dataId` munging like this shouldn't be needed at all . How does processCcd manage the case where 'hdu' is in the dataId?

kfindeisen: It doesn't do anything special (including when writing calexps).

As for the method as a whole being unnecessary, I think you're right. The original purpose (creating datarefs for calexp templates) would be better done using `Butler.subset` and a `run`-local butler.

kfindeisen: @parejkoj Looking again at the code, I think I originally introduced `_siblingRef` as an optimization for the following common case:

* The user specifies a (raw) data ID on the command line that restricts both visit and ccd, and
* The user specifies only a visit for `--templateId`, as [specified by `ImageDifferenceTask`](https://github.com/lsst/pipe_tasks/blob/master/python/lsst/pipe/tasks/imageDifference.py#L789-L790).

In fact, `GetCalexpAsTemplateTask` does exactly the same thing as `_siblingRef`: it [takes the primary dataRef and overwrites only its visit field](https://github.com/lsst/ip_diffim/blob/master/python/lsst/ip/diffim/getTemplate.py#L197-L198).

Anyway, I can replace the calls to `_siblingRef` with `butler.subset`, but this will mean that `ap_pipe` will always reduce an entire visit for `--templateId` even when only a few CCDs are needed.

parejkoj: Oof. This sort of dataId munging is probably going to break with Gen3 butler. I guess we can leave it as is and deal with it when the new API is available.

Please file a ticket about how to handle the "dataRef and templateId" question with gen3 butler (and note the similar uses, e.g. `GetCalexpAsTemplateTask`), and reference it in `_siblingRef`'s docstring (probably in `Notes`?). I think we need much better guidance on this sort of thing, but getting that guidance is at some point in the future.

kfindeisen: I'm pretty sure *all* butler usage will break in version 3.

Where exactly do you expect me to file a ticket? For which component, to do what?

PR #20, opened 2018-05-17
DM-14494: selectImages.py depends on defunct lsst.geom

kfindeisen:
This PR fixes an `ImportError` in `selectImages.py` that propagates to imports of `lsst.ap.pipe`. The `__init__` file for `ap.pipe` may also be edited as part of this PR, but I'm waiting for confirmation from @ebellm that it's safe to do so.

### DIFF
--- a/python/lsst/ap/pipe/selectImages.py
+++ b/python/lsst/ap/pipe/selectImages.py
@@ -28,7 +28,7 @@ import lsst.pex.exceptions as pexExceptions
 import lsst.afw.geom as afwGeom
 import lsst.pipe.base as pipeBase
 from lsst.pipe.tasks.selectImages import WcsSelectImagesTask, BaseExposureInfo
-from lsst.geom import convexHull
+from lsst.sphgeom import ConvexPolygon
 
 __all__ = ["MaxPsfWcsSelectImagesTask"]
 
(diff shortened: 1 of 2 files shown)


python/lsst/ap/pipe/selectImages.py
```diff
@@ -59,27 +59,27 @@ def runDataRef(self, dataRef, coordList, makeDataRefList=True, selectDataList=[]
...
         -----
-        We use the "convexHull" function in the geom package to define
+        We use the "convexHull" method in the `~lsst.sphgeom` package to define
         polygons on the celestial sphere, and test the polygon of the
         patch for overlap with the polygon of the image.
-
-        We use "convexHull" instead of generating a SphericalConvexPolygon
-        directly because the standard for the inputs to SphericalConvexPolygon
-        are pretty high and we don't want to be responsible for reaching them.
-        If "convexHull" is found to be too slow, we can revise this.
```
mrawls: I don't entirely understand the context for this original comment, but given that the entirety of `selectImages.py` will be removed as part of DM-11953, I'm not overly concerned about it.

kfindeisen: `lsst.geom` had a "with checks" and a "without checks" way to get a convex polygon. `lsst.sphgeom` always checks, as far as I can tell.

PR #21, opened 2018-05-18
DM-14497: ap_pipe doesn't know filters for AssociationTask

kfindeisen:
This PR updates `ApPipeConfig` to account for API changes in `AssociationTask`. It also ~adds a temporary workaround for DM-14507 (though I'm not sure why it works, since `AssociationTask` does the same thing internally) and~ improves the input validation of `ApPipeConfig`.

### DIFF
--- a/python/lsst/ap/pipe/ap_pipe.py
+++ b/python/lsst/ap/pipe/ap_pipe.py
@@ -104,6 +104,12 @@ class ApPipeConfig(pexConfig.Config):
 
     def validate(self):
         pexConfig.Config.validate(self)
+        if not self.differencer.doMeasurement:
+            raise ValueError("Source association needs diaSource fluxes.")
+        if not self.differencer.doWriteSources:
+            raise ValueError("Source association needs diaSource catalogs.")
+        if not self.differencer.doWriteSubtractedExp:
+            raise ValueError("Source association needs difference exposures.")
 
 
 class ApPipeTask(pipeBase.CmdLineTask):
--- a/python/lsst/ap/pipe/ap_pipe.py
+++ b/python/lsst/ap/pipe/ap_pipe.py
@@ -99,6 +99,8 @@ class ApPipeConfig(pexConfig.Config):
         self.differencer.doSelectSources = False
 
         self.associator.level1_db.retarget(AssociationDBSqliteTask)
+        # TODO: generalize in DM-12315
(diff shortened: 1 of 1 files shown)


python/lsst/ap/pipe/ap_pipe.py
```diff
@@ -99,9 +99,17 @@ def setDefaults(self):
         self.differencer.doSelectSources = False
 
         self.associator.level1_db.retarget(AssociationDBSqliteTask)
+        # TODO: generalize in DM-12315
+        self.associator.level1_db.filter_names = ['u', 'g', 'r', 'i', 'z', 'y', 'VR', 'N964']
```
morriscb: Why doesn't using a configfile for in the CmdLineTask work here? This looks very similar to how it would happen in such a file. I'm worried that this may be adding in values specific to DECam.

python/lsst/ap/pipe/ap_pipe.py
```diff
@@ -293,8 +301,13 @@ def runAssociation(self, sensorRef):
         self.log.info("Running Association...")
 
         diffType = self.config.differencer.coaddName
+        psfType = self.config.differencer.measurement.slots.psfFlux
         try:
             catalog = sensorRef.get(diffType + "Diff_diaSrc")
+            # TODO: workaround for DM-14507
```
morriscb: Now that we've found the bug in ap_association that was causing this do you want to wait for me to merge that ticket or do this for now and then remove it later?

python/lsst/ap/pipe/ap_pipe.py
kfindeisen: Almost everything in the current default config is DECam-specific; DM-12315 will fix that (see also the Jira description of DM-14497).

python/lsst/ap/pipe/ap_pipe.py
kfindeisen: Sure, I can wait.

python/lsst/ap/pipe/ap_pipe.py
morriscb: Sounds good.

python/lsst/ap/pipe/ap_pipe.py
morriscb: This is not fixed in DM-14517. Thanks for the help.

PR #26, opened 2018-07-16
DM-14259: Try and document running ap_pipe on the Verification Cluster with SLURM

mrawls:
The documentation of my workflow is in a comment on Jira.


bin/make_run_ap_pipe_conf.py
```diff
@@ -0,0 +1,24 @@
...
+          411456, 411470, 411472, 411657, 411671, 411673, 411707, 411721, 411724,
+          411758, 411772, 411774, 411808, 411822, 411824, 411858, 411872, 411874,
+          412060, 412074, 412076, 412250, 412264, 412266, 412307, 412321,
+          412324, 412504, 412518, 412520, 412554, 412568, 412570, 412604,
+          412618, 412620, 412654, 412668, 412670, 412704, 412718, 412720,
+          413635, 413649, 413651, 413680, 413694, 413696,
+          415314, 415328, 415330, 415364, 415378, 415380,
+          419791, 419802, 419804, 421590, 421604, 421606]
+
+filter = 'g'
```
ebellm: filter is a Python builtin function, so I'd suggest a different variable name here.

bin/make_run_ap_pipe_conf.py
```diff
@@ -0,0 +1,24 @@
+import os
+'''Generate configuration file for ap_pipe slurm run
+'''
+
+BASE_DIR = '/project/mrawls/hits2015/'
```
ebellm: Could we make this more generic, perhaps by taking BASE_DIR as an argv argument?

bin/run_ap_pipe.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+# Set up the stack and necessary ap_packages
+source /software/lsstsw/stack/loadLSST.bash
+setup lsst_distrib
+
+# Set up the cloned and pre-sconsed ap_packages
+L1DBPROTO_DIR=/project/mrawls/l1dbproto
```
ebellm: Suggest for all of these that you either allow them to be passed as arguments or use `os.getenv` to get environment variables.  I just think we'll quickly find it convenient to be able to run these scripts without editing them.

bin/run_ap_pipe.sh
```diff
@@ -0,0 +1,33 @@
...
+AP_ASSOC_DIR=/project/mrawls/ap_association
+AP_PIPE_DIR=/project/mrawls/ap_pipe
+setup -k -r $L1DBPROTO_DIR
+setup -k -r $AP_ASSOC_DIR
+setup -k -r $AP_PIPE_DIR
+
+# Set environment variables used to run ap_pipe.py on the command line
+REPO=/project/mrawls/hits2015/
```
ebellm: same comment about arguments/env variables

bin/make_run_ap_pipe_conf.py
```diff
@@ -0,0 +1,24 @@
+import os
+'''Generate configuration file for ap_pipe slurm run
```
ebellm: maybe the filename should make clear this is only for HiTS?  unless you want to make visits configurable in some way.

bin/run_ap_pipe.sh
```diff
@@ -0,0 +1,33 @@
...
+CALIB=/project/mrawls/hits2015/calib3
+TEMPLATE=/project/mrawls/hits2015/templates
+RERUN=slurm1
+
+# Run ap_pipe in a series of jobs split up by visit and ccdnum
+
+# Note CCDs 2 and 61 are missing, so we process 3-60 inclusive and then 1 and 62
```
ebellm: Again, the filename here says `run_ap_pipe`, but we're now restricted to DECam.  I'm not sure it's worth dealing with the overhead of passing general ccd ranges to this bash script (we don't have the stack tools to use), but maybe it should be called `run_ap_pipe_decam`?

bin/prep_ap_pipe.sh
```diff
@@ -0,0 +1,73 @@
+#!/bin/bash
+
+# A script to set things up to run ap_pipe via slurm on some DECam data.
+#
+# First, run this script:
+#       $ bash prep_ap_pipe.sh $1=rerun $2=repo $3=calib $4=template $5=filter
+# e.g., $ bash prep_ap_pipe.sh slurm1 . calibdir templatedir g
+#
+# (Note each argument is optional and is set with defaults below)
```
ebellm: Because you're using positional arguments, if you don't set one it will mess up the others: it's all or nothing.  This is also annoying because you can't just change one and keep the others as defaults, for instance.

One option would be to write a parser that let you give arguments like --repo x, etc. (see http://linuxcommand.org/lc3_wss0120.php).  Another option would be to declare environment variables outside the script (e.g., `export FILTERNAME=g`) and detect if they are set or unset.

bin/prep_ap_pipe.sh
```diff
@@ -0,0 +1,73 @@
...
+# The result of running this script is to create a .conf file
+# (Before we write to it, we remove any old one lying around)
+# Also create a directory to write slurm outfiles to if it doesn't exist yet
+CONFFILE=run_ap_pipe.conf
+rm -f ${CONFFILE}
+mkdir -p slurm
+
+# Create CONFFILE
+# Loop over DECam CCDs (note ccdnums 2 and 61 will fail)
+for ccdidx in {1..62};
```
ebellm: If you add a NCCDS argument above (defaulting to 62) this will then be generic to non-DECam cameras.

mrawls: Sadly not, because decam has the field "ccdnum" while other cameras use "ccd." (That's one reason I didn't bother doing this.)

ebellm: Sounds like it's time for another variable!

(This does make me think that doing this in stack-land may be easier eventually.  Or harmonizing our obs packages further...)

mrawls:
For reals this time.

PR #27, opened 2018-07-24
DM-12315: Generalize ap_pipe to non-HiTS data

kfindeisen:
This PR moves most of the config defaults out of the source code (see lsst/ap_verify_hits2015#9, lsst/obs_decam#88, and lsst/obs_subaru#143). Note that, depending on the templates or refcats being used, runs of `ap_pipe` are much more likely to need a user config file.

This PR also makes some minor maintenance updates to `ap_pipe`.

### DIFF
--- a/python/lsst/ap/pipe/ap_pipe.py
+++ b/python/lsst/ap/pipe/ap_pipe.py
@@ -23,16 +23,12 @@
 
 __all__ = ["ApPipeConfig", "ApPipeTask"]
 
-import os
-
 import lsst.pex.config as pexConfig
 import lsst.pipe.base as pipeBase
 
 from lsst.pipe.tasks.processCcd import ProcessCcdTask
-from lsst.meas.algorithms import LoadIndexedReferenceObjectsTask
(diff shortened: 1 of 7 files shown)


python/lsst/ap/pipe/ap_pipe.py
```diff
@@ -55,52 +51,14 @@ class ApPipeConfig(pexConfig.Config):
...
-        self.ccdProcessor.calibrate.photoRefObjLoader.ref_dataset_name = "pan-starrs"
-        self.ccdProcessor.calibrate.photoRefObjLoader.filterMap = {
-            "u": "g",
-            "g": "g",
-            "r": "r",
-            "i": "i",
-            "z": "z",
-            "y": "y",
-            "VR": "g"}
-
```
mrawls: It looks like you've moved all this refcat filterMap business into `config/apPipe.py` in `ap_verify_hits2015`. Do all our "datasets" now need to include a config like this in order to run ap_pipe on them? Could this information live in the obs_package's `config/apPipe.py` instead?

PR #29, opened 2018-07-31
DM-15096: Cannot reuse association database with ApPipeTask reruns

kfindeisen:
This PR removes the database-in-repository model from `ap_pipe` as well as all the code that was necessary to enforce it. SQLite configs are no longer allowed to specify `:memory:`, as testing revealed that in-memory databases are incompatible with the requirement that the database be fully initialized before `AssociationTask` is created. The documentation has been updated to emphasize that the user must always choose a database location in addition to the usual output repository.

### DIFF
--- a/doc/lsst.ap.pipe/pipeline-tutorial.rst
+++ b/doc/lsst.ap.pipe/pipeline-tutorial.rst
@@ -64,7 +64,7 @@ To process your ingested data, run
 
 .. prompt:: bash
    
-   ap_pipe.py repo --calib repo/calibs --rerun processed --id visit=123456 ccdnum=42 filter=g --template templates
+   ap_pipe.py repo --calib repo/calibs --rerun processed -c associator.level1_db.db_name=ppdb/association.db --id visit=123456 ccdnum=42 filter=g --template templates
 
 In this case, a ``processed`` directory will be created within
 ``repo/rerun`` and the results will be written there.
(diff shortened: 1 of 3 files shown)


doc/lsst.ap.pipe/pipeline-tutorial.rst
```diff
@@ -64,7 +64,7 @@ To process your ingested data, run
 
 .. prompt:: bash
    
-   ap_pipe.py repo --calib repo/calibs --rerun processed --id visit=123456 ccdnum=42 filter=g --template templates
+   ap_pipe.py repo --calib repo/calibs --rerun processed -c associator.level1_db.db_name=association.db --id visit=123456 ccdnum=42 filter=g --template templates
```
mrawls: I believe @morriscb said over on Jira that the config field should be `db_url`, not `db_name`. Unless I'm misinterpreting, please update this accordingly.

doc/lsst.ap.pipe/pipeline-tutorial.rst
```diff
@@ -86,12 +86,16 @@ If you prefer to have a standalone output repository, you may instead run
...
+   ap_pipe.py repo --calib repo/calibs --output path/to/put/processed/data/in -c associator.level1_db.db_name=association.db --id visit=123456 ccdnum=42 filter=g --template path/to/templates
 
 In this case, the output directory will be created if it does not already exist.
 If you omit the ``--template`` flag, ``ap_pipe`` will assume the templates are
 somewhere in ``repo``.
 
+.. note::
+
+   You must :ref:`configure <command-line-task-config-howto>` the association database location.
+   In the examples above, this is done with the ``-c`` option, but a personal config file may be more convenient if you intend to run ``ap_pipe`` many times.
```
mrawls: What happens if the user fails to configure the database location and tries to run ap_pipe anyway? Please add a sentence about what will happen in this situation.

doc/lsst.ap.pipe/pipeline-tutorial.rst
```diff
@@ -104,11 +108,11 @@ something like
 
 .. code-block:: none
 
+   association.db   <--- the Prompt Products Database with DIAObjects
    repo/
```
mrawls: What I take away from this is that the database with whatever name the user chooses will be created in the same directory which contains `repo`, assuming there is not already a file there with that name. Is that right?

doc/lsst.ap.pipe/pipeline-tutorial.rst
mrawls: Perhaps more importantly, it's not clear where (path?) this Must-Be-Named database will be created (or is expected to already exist). Please clarify this and make sure it is consistent with the pseudo file-tree in the Expected Outputs section. Can a path be given for the `db_name`/`db_url` or just a name?

You say the database name must be specified, but please explain what it does with the name. Something like, "ap_pipe will look for a database that already exists at the location specified. If it finds one, it will connect to it and continue associating, and if it doesn't find one, it will create one."

doc/lsst.ap.pipe/pipeline-tutorial.rst
kfindeisen: No, the config field is `db_name`. It we for some reason had to configure `l1dbproto` then that would be relevant, but AFAIK we don't. (I can't test it because `AssociationDBSqliteTask` and `AssociationL1DBProtoTask` have incompatible APIs, and `ap_pipe` assumes the former.)

doc/lsst.ap.pipe/pipeline-tutorial.rst
kfindeisen: No, it's created relative to the working directory. Is there some obvious other place to put it, given that spelling out a full rerun path kind of defeats the point of reruns?

kfindeisen: That sentence sounds too much like saying there's a default location. How about changing the example to `-c associator.level1_db.db_name=ppdb/association.db` and modifying the "after" directory layout to match? Will that make things unambiguous?

mrawls: Yes, that sounds good. I understand you want to avoid hinting at any kind of default. Thanks!

PR #31, opened 2018-09-28
DM-14273: Add ap_verify to lsst_distrib

kfindeisen:
This PR makes various package cleanups to conform to DM standards, and updates the documentation to assume `ap_pipe` is available from `lsst_apps`.


ups/ap_pipe.table
```diff
@@ -10,7 +10,6 @@ setupRequired(utils >= 4.6.0.0)
 setupRequired(pex_exceptions >= 4.6.0.0)
 setupRequired(ndarray)
 
-setupRequired(obs_decam)
```
timj: I don't think you depend on ndarray, pex_exceptions, or boost either so please remove them.

doc/lsst.ap.pipe/getting-started.rst
```diff
@@ -10,16 +10,8 @@ Getting started with the AP Pipeline
...
-You will need to clone, setup, and build it by following the
-`Science Pipelines documentation <https://pipelines.lsst.io/install/package-development.html>`_.
-
-.. note::
-
-   `lsst.ap.association`, available `here <https://github.com/lsst-dm/ap_association>`_,
-   is a prerequisite for `lsst.ap.pipe` and is also not currently part of the
-   LSST Stack. You must clone ``ap_association`` and set it up first.
+`lsst.ap.pipe` is available from the `LSST Science Pipelines <https://pipelines.lsst.io/>`_.
+It is installed as part of the ``lsst_apps`` and ``lsst_distrib`` metapackages.
```
timj: It's being added to lsst_apps, not really lsst_distrib whereas this sentence implies it's part of both (distrib installs apps which happens to include ap_pipe).

kfindeisen: Well, it *is* part of both, and deliberately so.

I deliberately worded it this way because the installation instructions are geared towards `ap_pipe`'s current user base, who are not necessarily familiar with the Stack. It would not be obvious to them that installing `lsst_distrib` gets you everything in `lsst_apps`, and in practice current users will want to install `lsst_distrib` anyway for the `obs` packages.

PR #34, opened 2018-11-19
DM-15588: Remove home-brewed SQLite PPDB

morriscb:


python/lsst/ap/pipe/ap_pipe.py
```diff
@@ -245,17 +276,15 @@ def runAssociation(self, sensorRef):
...
-            result = self.associator.run(catalog, exposure)
-        finally:
-            # Stateful AssociationTask will work for now because TaskRunner
-            # uses task-oriented parallelism. Will not be necessary after DM-13672
-            self.associator.level1_db.close()
+
+        catalog = sensorRef.get(diffType + "Diff_diaSrc")
+        exposure = sensorRef.get(diffType + "Diff_differenceExp")
+
+        dia_sources = self.diaSourceDpddifier.run(catalog, exposure)
```
kfindeisen: I assume (hope) the Dpddifier is a temporary measure; is there some ticket we can `TODO` here?

python/lsst/ap/pipe/ap_pipe.py
```diff
@@ -245,17 +276,15 @@ def runAssociation(self, sensorRef):
...
+        result = self.associator.run(dia_sources, exposure, self.ppdb)
 
         return pipeBase.Struct(
-            l1Database=self.associator.level1_db,
+            l1Database=self.ppdb,
```
kfindeisen: Please update the documentation for `runAssociation`'s return value.

python/lsst/ap/pipe/ap_pipe.py
```diff
@@ -278,12 +307,16 @@ def _setupDatabase(configurable):
         A ConfigurableInstance with a database-managing class in its ``target``
         field. The API of ``target`` must expose a ``create_tables`` method
         taking no arguments.
```
kfindeisen: This documentation is out of date. See my proposed change to `ApPipeConfig.ppdb` before updating, though.

python/lsst/ap/pipe/ap_pipe.py
```diff
@@ -278,12 +307,16 @@ def _setupDatabase(configurable):
...
     """
-    db = configurable.apply()
-    try:
-        db.create_tables()
-    finally:
-        db.close()
+    ppdb = daxPpdb.Ppdb(config=configurable,
+                        afw_schemas=dict(DiaObject=make_dia_object_schema(),
+                                         DiaSource=make_dia_source_schema()))
+    ppdb._schema.makeSchema()
```
kfindeisen: Given the discussion on lsst/dax_ppdb#5, this line should be `ppdb.makeSchema()` but otherwise doesn't need to be changed.

python/lsst/ap/pipe/ap_pipe.py
```diff
@@ -45,6 +53,15 @@ class ApPipeConfig(pexConfig.Config):
         target=ImageDifferenceTask,
         doc="Task used to do image subtraction and DiaSource detection.",
     )
+    ppdb = pexConfig.ConfigField(
+        dtype=daxPpdb.PpdbConfig,
+        doc="Ppdb database configuration.",
+    )
```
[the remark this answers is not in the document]
morriscb: I think the problem is that Ppdb isn't a task (it's base class is object) so I can't use the config.apply() I think. Correct me if I'm wrong though.

python/lsst/ap/pipe/ap_pipe.py
morriscb: So I should do both correct? (add the comment and change it to makeSchema).

python/lsst/ap/pipe/ap_pipe.py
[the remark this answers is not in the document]
kfindeisen: I haven't tested it, but the [documentation](http://doxygen.lsst.codes/stack/doxygen/x_masterDoxyDoc/classlsst_1_1pex_1_1config_1_1configurable_field_1_1_configurable_field.html#details) doesn't say anything about the target having to be a task. It should work on any class that takes a config as its first constructor parameter:
> Must be callable, and the first parameter will be the value of this field

Targets usually have a `ConfigClass` member (which `Ppdb` doesn't), but you can override that when creating the field.

python/lsst/ap/pipe/ap_pipe.py
morriscb: Not currently. Currently the DPDD-ification process Yusra is working on is still a prototype. At some point we may switch over to that but there will always be some process in between the output of ip_diffim and ap_association/pipe.

python/lsst/ap/pipe/ap_pipe.py
```diff
@@ -69,11 +100,9 @@ def validate(self):
         if not self.differencer.doWriteSubtractedExp:
             raise ValueError("Source association needs difference exposures "
                              "[differencer.doWriteSubtractedExp].")
-
-        # Not all level1_db implementations let you specify a DB location
-        if hasattr(self.associator.level1_db, "db_name") and self.associator.level1_db.db_name == ":memory:":
-            raise ValueError("Source association needs a persistent database [associator.level1_db.db_name]. "
-                             "Please provide a DB file (need not exist) or use a different level1_db Task.")
+        if self.ppdb.isolation_level == "READ_COMMITTED" and self.ppdb.db_url[:6] == "sqlite":
```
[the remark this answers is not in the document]
morriscb: Added to Ppdb instead.

python/lsst/ap/pipe/ap_pipe.py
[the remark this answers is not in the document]
morriscb: Okay, so it worked. The apply function seems to do what is expected. I've moved the ppdb.makeSchema() call into __init__ and deleted the "_connectToDB" as it was now completely handled by the `apply` call.

python/lsst/ap/pipe/ap_pipe.py
morriscb: Now that the `config.apply` method I've removed this part of the code.

python/lsst/ap/pipe/ap_pipe.py
```diff
@@ -60,6 +77,20 @@ def setDefaults(self):
...
+        self.ppdb.schema_file = os.path.join(
+            getPackageDir("dax_ppdb"), "data", "ppdb-schema.yaml")
+        self.ppdb.column_map = os.path.join(
+            getPackageDir("ap_association"),
+            "data",
+            "ppdb-ap-pipe-afw-map.yaml")
+        self.ppdb.extra_schema_file = os.path.join(
+            getPackageDir("ap_association"),
+            "data",
+            "ppdb-ap-pipe-schema-extra.yaml")
```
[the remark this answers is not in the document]
morriscb: Removed some of the superfluous settings as the schema for Association has moved closer to that of the default Ppdb and most of the defaults are held there. Once we need to start using the fancier features of the Ppdb I'll consider using ap_association specific config.

PR #40, opened 2019-05-03
DM-13567: Add tests to ap_pipe

mrawls:
Pair coded with @kfindeisen. The new tests require the ap_pipe_testdata package.

### DIFF
--- a/tests/test_appipe.py
+++ b/tests/test_appipe.py
@@ -21,22 +21,152 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 
+import contextlib
+import os
 import unittest
+from unittest.mock import patch, Mock
+
 import lsst.utils.tests
+import lsst.pex.exceptions as pexExcept
+import lsst.daf.persistence as dafPersist
+import lsst.pipe.base as pipeBase
+
+from lsst.ap.pipe import ApPipeTask
 
 
 class PipelineTestSuite(lsst.utils.tests.TestCase):
     '''
     A set of tests for the functions in ap_pipe.
-
-    TODO: write tests for DM-13567.
     '''
 
-#    def testDoProcessCcd(self):
-#        # test something
+    @classmethod
+    def _makeDefaultConfig(cls):
+        config = ApPipeTask.ConfigClass()
+        config.load(os.path.join(cls.datadir, "config", "apPipe.py"))
+        config.ppdb.db_url = "sqlite://"
+        config.ppdb.isolation_level = "READ_UNCOMMITTED"
+        return config
 
-#    def testDoDiffIm(self):
-#        # test something
+    @classmethod
+    def setUpClass(cls):
+        try:
+            cls.datadir = lsst.utils.getPackageDir("ap_pipe_testdata")
+        except pexExcept.NotFoundError:
+            raise unittest.SkipTest("ap_pipe_testdata not set up")
+
+    def setUp(self):
+        self.config = self._makeDefaultConfig()
+        self.butler = dafPersist.Butler(inputs={'root': self.datadir})
+
+        def makeMockDataRef(datasetType, level=None, dataId={}, **rest):
+            mockDataRef = Mock(dafPersist.ButlerDataRef)
+            mockDataRef.dataId = dict(dataId, **rest)
+            return mockDataRef
+
+        butlerPatcher = patch.object(self.butler, "dataRef",
+                                     side_effect=makeMockDataRef)
+        butlerPatcher.start()
+        self.addCleanup(butlerPatcher.stop)
+        self.dataId = {"visit": 413635, "ccdnum": 42}
+        self.inputRef = self.butler.dataRef("raw", **self.dataId)
+
+    @contextlib.contextmanager
+    def mockPatchSubtasks(self, task):
+        """Make mocks for all the ap_pipe subtasks.
+
+        This is needed because the task itself cannot be a mock.
+        The task's subtasks do not exist until the task is created, so
+        this allows us to mock them instead.
+
+        Parameters
+        ----------
+        task : `lsst.ap.pipe.ApPipeTask`
+            The task whose subtasks will be mocked.
+
+        Yields
+        ------
+        subtasks : `lsst.pipe.base.Struct`
+            All mocks created by this context manager, including:
+
+            ``ccdProcessor``
+            ``differencer``
+            ``dpddifier``
+            ``associator``
+            ``forcedSource``
+                a mock for the corresponding subtask. Mocks do not return any
+                particular value, but have mocked methods that can be queried
+                for calls by ApPipeTask
+        """
+        with patch.object(task, "ccdProcessor") as mockCcdProcessor, \
+                patch.object(task, "differencer") as mockDifferencer, \
+                patch.object(task, "diaSourceDpddifier") as mockDpddifier, \
+                patch.object(task, "associator") as mockAssociator, \
+                patch.object(task, "diaForcedSource") as mockForcedSource:
+            yield pipeBase.Struct(ccdProcessor=mockCcdProcessor,
+                                  differencer=mockDifferencer,
+                                  dpddifier=mockDpddifier,
+                                  associator=mockAssociator,
+                                  forcedSource=mockForcedSource)
+
+    def testGenericRun(self):
+        """Test the normal workflow of each ap_pipe step.
+        """
+        task = ApPipeTask(self.butler, config=self.config)
+        with self.mockPatchSubtasks(task) as subtasks:
+            task.runDataRef(self.inputRef)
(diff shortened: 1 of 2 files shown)


tests/test_appipe.py
```diff
@@ -21,22 +21,142 @@
...
+        return config
+
+    @classmethod
+    def setUpClass(cls):
+        try:
+            cls.datadir = lsst.utils.getPackageDir("ap_pipe_testdata")
+        except pexExcept.NotFoundError:
+            raise unittest.SkipTest("ap_pipe_testdata not set up")
+
+        cls.config = cls._makeDefaultConfig()
```
parejkoj: I would have `config` be set in `setUp`, so that it is always reset for each test.

tests/test_appipe.py
```diff
@@ -21,22 +21,142 @@
...
+
+    def setUp(self):
+        self.butler = dafPersist.Butler(inputs={'root': self.datadir})
+
+        def makeMockDataRef(datasetType, level=None, dataId={}, **rest):
+            mockDataRef = Mock(dafPersist.ButlerDataRef)
+            mockDataRef.dataId = dict(dataId, **rest)
```
parejkoj: I'm surprised that this worked! Guess I learned a new thing about `dict` initialization today.

tests/test_appipe.py
```diff
@@ -21,22 +21,142 @@
...
+            ``forcedSource``
+                a mock for the corresponding subtask. Mocks do not return any
+                particular value, but have mocked methods that can be queried
+                for calls by ApPipeTask
+        """
+        with patch.object(task, "ccdProcessor") as mockCcdProcessor:
+            with patch.object(task, "differencer") as mockDifferencer:
+                with patch.object(task, "diaSourceDpddifier") as mockDpddifier:
+                    with patch.object(task, "associator") as mockAssociator:
+                        with patch.object(task, "diaForcedSource") as mockForcedSource:
```
parejkoj: Ouch. You don't need to nest these, you can put them all in one context: `with patch1() as mock1, patch2() as mock2, ...: doStuff()`.

tests/test_appipe.py
```diff
@@ -21,22 +21,142 @@
...
+                            yield pipeBase.Struct(ccdProcessor=mockCcdProcessor,
+                                                  differencer=mockDifferencer,
+                                                  dpddifier=mockDpddifier,
+                                                  associator=mockAssociator,
+                                                  forcedSource=mockForcedSource)
+
+    def testGenericRun(self):
+        # test the normal workflow of each ap_pipe step
+        task = ApPipeTask(self.butler, config=self.config)
+        inputRef = self.butler.dataRef("raw", **self.dataId)
```
parejkoj: I see this `inputRef` repeated below: can you put it into `setUp`?

tests/test_appipe.py
```diff
@@ -21,22 +21,142 @@
...
+                    with patch.object(task, "associator") as mockAssociator:
+                        with patch.object(task, "diaForcedSource") as mockForcedSource:
+                            yield pipeBase.Struct(ccdProcessor=mockCcdProcessor,
+                                                  differencer=mockDifferencer,
+                                                  dpddifier=mockDpddifier,
+                                                  associator=mockAssociator,
+                                                  forcedSource=mockForcedSource)
+
+    def testGenericRun(self):
+        # test the normal workflow of each ap_pipe step
```
parejkoj: Comment should be docstring?

tests/test_appipe.py
```diff
@@ -21,22 +21,142 @@
...
+        inputRef = self.butler.dataRef("raw", **self.dataId)
+        with self.mockPatchSubtasks(task) as subtasks:
+            task.runDataRef(inputRef)
+            subtasks.ccdProcessor.runDataRef.assert_called_once()
+            subtasks.differencer.runDataRef.assert_called_once()
+            subtasks.associator.run.assert_called_once()
+            subtasks.forcedSource.run.assert_called_once()
+
+    def testReuseExistingOutput(self):
+        # test reuse keyword to ApPipeTask.runDataRef
```
parejkoj: Comment should be docstring?

tests/test_appipe.py
```diff
@@ -21,22 +21,142 @@
...
+        task = ApPipeTask(self.butler, config=self.config)
+        inputRef = self.butler.dataRef("raw", **self.dataId)
+
+        # Mock guarantees that all "has this been made" tests pass,
+        # so skippable subtasks should actually be skipped
+        for skippable in [
+            ['ccdProcessor'],
+            ['ccdProcessor', 'differencer'],
+        ]:
```
parejkoj: Loops in tests make the tests difficult to untangle. You're only doing two things here, so I think it would be clearer if you made the internals of this into a separate method (e.g. `checkSkippedTask`) and then call it:

```
self.checkSkippedTask(['ccdProcessor'])
self.checkSkippedTask(['ccdProcessor', 'differencer'])
```

Similarly for the loop below, although I'm not sure how to unravel that one.

tests/test_appipe.py
```diff
@@ -21,22 +21,142 @@
...
+        calexpConfig.differencer.doSelectSources = False  # Workaround for DM-18394
 
-#    def testDoDiffIm(self):
-#        # test something
+        task = ApPipeTask(self.butler, config=calexpConfig)
+        inputRef = self.butler.dataRef("raw", **self.dataId)
+        with self.mockPatchSubtasks(task) as subtasks:
+            # Self-subtraction ok because we're not actually doing any processing
+            task.runDataRef(inputRef, templateIds=[self.dataId])
+            subtasks.ccdProcessor.runDataRef.assert_called()  # Should be twice
```
parejkoj: You can check `assertEqual(runDataRef.call_count, 2)` here.

tests/test_appipe.py
```diff
@@ -21,22 +21,142 @@
...
+        calexpConfig = self._makeDefaultConfig()
+        calexpConfig.load(calexpConfigFile)
+        calexpConfig.differencer.doSelectSources = False  # Workaround for DM-18394
 
-#    def testDoDiffIm(self):
-#        # test something
+        task = ApPipeTask(self.butler, config=calexpConfig)
+        inputRef = self.butler.dataRef("raw", **self.dataId)
+        with self.mockPatchSubtasks(task) as subtasks:
+            # Self-subtraction ok because we're not actually doing any processing
```
parejkoj: I don't understand what this comment is referring to?

tests/test_appipe.py
```diff
@@ -21,22 +21,142 @@
...
+            return mockDataRef
+        butlerPatcher = patch.object(self.butler, "dataRef",
+                                     side_effect=makeMockDataRef)
+        butlerPatcher.start()
+        self.addCleanup(butlerPatcher.stop)
+        self.dataId = {"visit": 413635, "ccdnum": 42}
+
+    @contextlib.contextmanager
+    def mockPatchSubtasks(self, task):
+        """Make mocks for all the ap_pipe subtasks.
```
parejkoj: I think it's worth explaining in the docstring why this context manager is necessary: I was thinking that it wasn't because calls to mocks always return other mocks, so e.g. `task.differencer.runDataRef` is a mock. Except that `task` is not a mock, so you do need this. I thought you might be able to remove the context manager and put it into setup, patching `ApPipeTask` itself, but the subtasks don't exist until the Task is created, so you do need this.

So yes, some more explanation here would be good.

tests/test_appipe.py
```diff
@@ -21,22 +21,142 @@
...
+
+        Parameters
+        ----------
+        task : `lsst.ap.pipe.ApPipeTask`
+            the task whose subtasks will be mocked
```
parejkoj: Sentence.

tests/test_appipe.py
mrawls: Thank you, that works!

tests/test_appipe.py
mrawls: We fixed the first loop as suggested and added a message that identifies the subtask being tested for the second loop.

tests/test_appipe.py
mrawls: Clarified. It was referring to the fact that the differencing template has the same dataId as the science template.

tests/test_appipe.py
```diff
@@ -21,22 +21,149 @@
...
+    def testReuseExistingOutput(self):
+        """Test reuse keyword to ApPipeTask.runDataRef.
+        """
+        task = ApPipeTask(self.butler, config=self.config)
+
+        self.checkReuseExistingOutput(task, ['ccdProcessor'])
+        self.checkReuseExistingOutput(task, ['ccdProcessor', 'differencer'])
+
+    def checkReuseExistingOutput(self, task, skippable):
+        # Mock guarantees that all "has this been made" tests pass,
```
parejkoj: Docstring for this method?

PR #42, opened 2019-06-05
DM-18736: Convert ap_association to use Pandas data frames (rather than afw::table) as an interface

morriscb:
Comment out diaForcedSourceTask for now.

### DIFF
--- a/python/lsst/ap/pipe/ap_pipe.py
+++ b/python/lsst/ap/pipe/ap_pipe.py
@@ -300,12 +300,17 @@ class ApPipeTask(pipeBase.CmdLineTask):
         catalog = sensorRef.get(diffType + "Diff_diaSrc")
         diffim = sensorRef.get(diffType + "Diff_differenceExp")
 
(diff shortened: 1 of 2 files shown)


python/lsst/ap/pipe/ap_pipe.py
```diff
@@ -300,12 +300,17 @@ def runAssociation(self, sensorRef):
...
-        dia_sources = self.diaSourceDpddifier.run(catalog, diffim)
+        dia_sources = self.diaSourceDpddifier.run(catalog,
+                                                  diffim,
+                                                  return_pandas=True)
         result = self.associator.run(dia_sources, diffim, self.ppdb)
-        self.diaForcedSource.run(result.dia_objects,
-                                 sensorRef.get("ccdExposureId_bits"),
-                                 sensorRef.get("calexp"),
-                                 diffim)
+        # TODO: DM-19960
```
ebellm: This appears to be the wrong ticket number.

PR #48, opened 2019-12-10
DM-22478: Lightly restructure ap_association tasks and create new DiaObject/DiaSource pre-loading task.

morriscb:
Modify runAssociation for new method inputs.

Fix bug in runAssociation

Add validation test for HTM level.

Debug validation.

### DIFF
--- a/python/lsst/ap/pipe/ap_pipe.py
+++ b/python/lsst/ap/pipe/ap_pipe.py
@@ -37,6 +37,7 @@ from lsst.pipe.tasks.imageDifference import ImageDifferenceTask
 from lsst.ap.association import (
     AssociationTask,
     DiaForcedSourceTask,
+    LoadDiaCatalogsTask,
     MapDiaSourceTask,
     make_dia_object_schema,
     make_dia_source_schema)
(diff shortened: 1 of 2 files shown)


python/lsst/ap/pipe/ap_pipe.py
```diff
@@ -105,6 +110,10 @@ def validate(self):
         if not self.differencer.doWriteSubtractedExp:
             raise ValueError("Source association needs difference exposures "
                              "[differencer.doWriteSubtractedExp].")
+        if self.diaCatalogLoader.htmLevel != \
+           self.associator.diaCalculation.plugins["ap_HTMIndex"].htmLevel:
+            raise ValueError("HTM index level in LoadDiaCatalogsTask must be equal to "
+                             "HTMIndexDiaCalculationPlugin index level.")
```
kfindeisen: :+1: for more validation, but does this need a check for whether the plugin is activated?

python/lsst/ap/pipe/ap_pipe.py
```diff
@@ -302,17 +312,23 @@ def runAssociation(self, sensorRef):
...
         diffim = sensorRef.get(diffType + "Diff_differenceExp")
 
-        dia_sources = self.diaSourceDpddifier.run(catalog,
-                                                  diffim,
-                                                  return_pandas=True)
+        diaSources = self.diaSourceDpddifier.run(catalog,
+                                                 diffim,
+                                                 return_pandas=True)
+        # Load the DiaObjects and DiaSource history.
+        loaderResult = self.diaCatalogLoader.run(diffim, self.apdb)
```
kfindeisen: I guess this line is outside the `try` block because if loading fails, then something's wrong with the DB and we can't do anything with it anyway? Might be worth a comment.

python/lsst/ap/pipe/ap_pipe.py
morriscb: Not here, but certainly within `AssociationTask` as none of the loading or storing in the Apdb will work without it. Added the validation to `AssociationTask`

PR #52, opened 2020-04-14
DM-24259: Create “stub“ Gen2 HSC dataset for CI testing

kfindeisen:
This PR adds a workaround for [DM-24435](https://jira.lsstcorp.org/browse/DM-24435), which otherwise causes `ap_verify` to crash when run with standard HSC configs.

This PR may not need to be merged; see lsst/pex_config#55.


kfindeisen:
The other PR seems to be delayed, so I've gone ahead and merged the workaround.

PR #58, opened 2020-08-31
DM-25748: Create pipeline task to generate fakes for AP processing and completeness tests

morriscb:


python/lsst/ap/pipe/createApFakes.py
```diff
@@ -0,0 +1,300 @@
...
+        Parameters
+        ----------
+        tractId : `int`
+            Tract id to produce randoms over.
+        skyMap : `lsst.skymap.SkyMap`
+            Skymap to produce randoms over.
+
+        Returns
+        -------
+        randoms : `lsst.pipe.tasks.PaquertTable`
```
kgabor, line 145: `PaquertTable` -> `ParquetTable`

python/lsst/ap/pipe/createApFakes.py
```diff
@@ -0,0 +1,300 @@
...
+            Catalog of random points covering the given tract. Follows the
+            format expected in `lsst.pipe.tasks.InsertFakes`.
+        """
+        np.random.seed(self.config.randomSeed)
```
kgabor, line 149: Use a np.random generator object instance instead of the built in random state. `rnd_gen = np.random.default_rng(seed)` etc. Add a note to the seed docstring that this goes to the default_rng().

https://numpy.org/doc/stable/reference/random/index.html

python/lsst/ap/pipe/createApFakes.py
```diff
@@ -0,0 +1,300 @@
...
+
+    def createRandomPositions(self, nFakes, boundingCircle):
+        """Create a set of spatially uniform randoms over the tract bounding
+        circle.
+
+        Parameters
+        ----------
+        nFakes : `int`
+            Number of fakes to create.
+        boundingCicle : `lsst.sphgeom.BoundingCircle`
```
kgabor, line 188: As far as I can see, all the random generation and transformation support up to a full sphere (up to Pi opening angle). Be explicit in the docstring about the maximum size of the allowed `BoundingCircle`.

python/lsst/ap/pipe/createApFakes.py
```diff
@@ -0,0 +1,300 @@
...
+            Circle bound covering the tract.
+
```
kgabor, line 190: Having a geometrical distribution of points can easily be ambiguous. Please add a Note in the docstring here and perhaps in the Task description as well that "spatially uniform" actually means a uniform distribution of points per solid angle here.

python/lsst/ap/pipe/createApFakes.py
```diff
@@ -0,0 +1,300 @@
...
+            the circle bound.
+        """
+        center = boundingCircle.getCenter()
+        cross = np.array([-center.y(),
+                          center.x(),
+                          0])
+        cross /= np.sqrt(cross[0] ** 2 + cross[1] ** 2 + cross[2] ** 2)
+        cosTheta = center.z()
+        sinTheta = np.sin(np.arccos(center.z()))
+
```
kgabor, line 247: I did not check that all the terms are correct in this matrix though I see it is tested with random angles (up to a rotation of ?); add a reference in a comment/docstring if this form of the rotation matrix is readily written down somewhere as a formula (I mean textbook, paper, WolframMathWorld, your notebook, tech note etc.).

python/lsst/ap/pipe/createApFakes.py
```diff
@@ -0,0 +1,300 @@
...
+            isVisitSource[:nBoth] = True
+            isTemplateSource[:nBoth] = True
+        if nOnly > 0:
+            isVisitSource[nBoth:(nBoth + nOnly)] = True
+            isTemplateSource[(nBoth + nOnly):] = True
+
+        return {self.config.visitSourceFlagCol: isVisitSource,
+                self.config.templateSourceFlagCol: isTemplateSource}
+
+    def createRandomMagnitudes(self, nFakes):
```
kgabor, line 294: Please emhpasize here (`createUniformRandomMagnitudes()`) and note in the Task docstring that this is _uniform in magnitudes_ which is a non-physical model but perhaps all right for this task.

kgabor (approved):

python/lsst/ap/pipe/createApFakes.py
morriscb: A larger than Pi value for a bounding circle doesn't make sense as a bounding circle of Pi is already encompasses the full sphere.

python/lsst/ap/pipe/createApFakes.py
morriscb: Wikipedia baby... https://en.wikipedia.org/wiki/Rotation_matrix the "Rotation from axis and angle" sub heading.

python/lsst/ap/pipe/createApFakes.py
morriscb: I kind of want to leave this as a generic name so that it can be overwritten in possible inherited tasks.

python/lsst/ap/pipe/createApFakes.py
morriscb: It's actually returning DataFrame now so I will change it.

python/lsst/ap/pipe/createApFakes.py
morriscb: added "on the sphere" to be clearer.

PR #59, opened 2020-09-16
DM-26553: Create gen3 unittests and test CreateApFakes pipeline tasks on data.

morriscb:


tests/test_createApFakes.py
```diff
@@ -54,6 +58,43 @@ def setUp(self):
...
+        dataId = {"skymap": "deepCoadd_skyMap", "tract": 0}
+        butler = butlerTests.makeTestCollection(testRepo)
+        butler.put(self.simpleMap, "deepCoadd_skyMap", {"skymap": "deepCoadd_skyMap"})
+
+        quantum = testUtils.makeQuantum(
+            fakesTask, butler, dataId,
+            {key: dataId for key in {"skyMap", "fakeCat"}})
+        run = testUtils.runTestQuantum(fakesTask, butler, quantum, True)
+        # Actual input dataset omitted for simplicity
+        run.assert_called_once()
```
kfindeisen, line 93: I skipped this in the example because it would have added a lot of distracting boilerplate, but since you have `self.simpleMap` and the tract ID, you should be able to predict the inputs to `run` and call `assert_called_once_with`.

tests/test_createApFakes.py
```diff
@@ -54,6 +58,43 @@ def setUp(self):
...
+        # Smoke test that the full pipeline task actualy runs.
+        testUtils.runTestQuantum(fakesTask, butler, quantum, False)
+        shutil.rmtree(root, ignore_errors=True)
+
```
kfindeisen, line 97: Seems redundant with `testRun`, but ok...

kfindeisen (approved):
Looks good, minor usage comments.

kfindeisen:
Actually, I do have one question that's outside the scope of the review: it looks like `CreateRandomApFakesTask` generates a fake source catalog given any skymap and tract. Aside from being useful to us, what does it have to do with AP or `ap_pipe`? Should it go into `pipe_tasks` instead, perhaps with a more generic name?

tests/test_createApFakes.py
morriscb: I think I can remove this since my concern here was that the values in runQuantum weren't being extracted properly. The `assert_called_once_with` resolves this I think.

morriscb:
There is some specificity here in the sense of the split between calexp and coadd source creation. The idea being they will show up in the diffIm differently. This config can be set up to pull all sources into both or you could just ignore the columns when running the insert tasks. I'm happy to leave this here for now and move it later. If @yalsayyad or @sr525 want, I can create a ticket to move this task into pipe_tasks so it can be used more generally down the road. I did code it as a kind of simple template class for creating fakes so it shouldn't be too hard to move it in the future.

PR #61, opened 2020-10-21
DM-27140 Update ApPipe to reflect changes to DRP pipeline

natelust:


pipelines/ApPipe.yaml
```diff
@@ -2,7 +2,9 @@
 
 description: End to end AP pipeline
 inherits:
-  - location: $PIPE_TASKS_DIR/pipelines/ProcessCcd.yaml
+  - location: $PIPE_TASKS_DIR/pipelines/DRP.yaml
+    include:
+      - processCcd
```
kfindeisen, line 7: It seems a bit odd for AP to delegate to DRP when they're supposed to be semi-independent pipelines, but if I understand subsets correctly, this does the job.

kfindeisen (approved):

PR #62, opened 2020-10-29
DM-22663: Reimplement make_apdb.py for Gen 3

kfindeisen:
This PR removes the `make_apdb.py` script's dependence on `pipe_base` and on `ApPipeConfig`, making it forward-compatible with Gen 3. As a side effect, the script now loads configs at a lower level, requiring changes to calling scripts.

### DIFF
--- a/doc/lsst.ap.pipe/apdb.rst
+++ b/doc/lsst.ap.pipe/apdb.rst
@@ -28,14 +28,13 @@ Configuring the database
 ========================
 
 The database is configured using `~lsst.dax.apdb.ApdbConfig`.
-|ap_pipe| command line users can pass configuration information to the script through the :option:`--config <ap_pipe.py --config>` and :option:`--configfile <ap_pipe.py --configfile>` command-line options.
-|make_apdb| also uses `ApPipeConfig` and the :option:`--config` and :option:`--configfile` options, so users can pass exactly the same arguments to |make_apdb| and |ap_pipe|.
-Supporting identical command line arguments for both scripts makes it easy to keep the database settings in sync.
+|ap_pipe| command line users can pass configuration information to the script through the :option:`--config <ap_pipe.py --config>` and :option:`--configfile <ap_pipe.py --configfile>` command-line options, using the prefix ``diaPipe.apdb.`` to distinguish APDB information from other pipeline configuration.
+|make_apdb| configures the database directly through :option:`--config` and :option:`--config-file` (different spelling, for consistency with |pipetask|), with no prefix.
 
-For |pipetask| users the process is almost the same, except that |pipetask|'s config syntax is not exactly the same.
-The :option:`--config <pipetask --config>` and :option:`--configfile <pipetask --configfile>` options for |pipetask| use colons as separators between each task and its config; replace these with periods for |make_apdb|.
+For |pipetask| users, the APDB is configured with the :option:`--config <pipetask run --config>` and :option:`--config-file <pipetask run --config-file>` options.
+APDB configuration info uses the prefix ``diaPipe:apdb.``, with a colon, but is otherwise the same.
 
-Note that ``apdb.db_url`` has no default; a value *must* be provided by the user.
+Note that the `~lsst.dax.apdb.ApdbConfig.db_url` field has no default; a value *must* be provided by the user.
 
 .. _section-ap-pipe-apdb-examples:
 
@@ -46,23 +45,35 @@ Databases can be configured using direct config overrides (see :ref:`ap-pipe-pip
 
 .. prompt:: bash
 
-   make_apdb.py -c diaPipe.apdb.isolation_level=READ_UNCOMMITTED diaPipe.apdb.db_url="sqlite:///databases/apdb.db" differencer.coaddName=dcr
+   make_apdb.py -c isolation_level=READ_UNCOMMITTED db_url="sqlite:///databases/apdb.db"
    ap_pipe.py -c diaPipe.apdb.isolation_level=READ_UNCOMMITTED diaPipe.apdb.db_url="sqlite:///databases/apdb.db" differencer.coaddName=dcr repo --calib repo/calibs --rerun myrun --id [optional IDs to process]
 
-|make_apdb| ignores any `ApPipeConfig` fields not related to the APDB (in the example, ``differencer.coaddName``), so there is no need to filter them out.
+The user is responsible for making sure the two APDB configurations are consistent.
 
 In Gen 3, this becomes (see :ref:`ap-pipe-pipeline-tutorial` for an explanation of |pipetask|):
 
 .. prompt:: bash
 
-   make_apdb.py -c diaPipe.apdb.isolation_level=READ_UNCOMMITTED diaPipe.apdb.db_url="sqlite:///databases/apdb.db" differencer.coaddName=dcr
+   make_apdb.py -c isolation_level=READ_UNCOMMITTED db_url="sqlite:///databases/apdb.db"
    pipetask run -p ApPipe.yaml -c diaPipe:apdb.isolation_level=READ_UNCOMMITTED diaPipe:apdb.db_url="sqlite:///databases/apdb.db" differencer:coaddName=dcr -b repo -o myrun
 
-Databases can also be set up using config files:
+.. warning::
+
+   As in Gen 2, make sure the APDB is created with a configuration consistent with the one used by the pipeline.
+   Note that the pipeline file given by ``-p`` may include APDB config overrides of its own.
+   You can double-check what configuration is being run by calling :command:`pipetask run` with the ``--show config="apdb*"`` argument, though this lists *all* configuration options, including those left at their defaults.
+
+Databases can also be set up using :ref:`config files <command-line-task-config-howto-configfile>`:
+
+.. code-block:: py
+   :caption: myApdbConfig.py
+
+   config.db_url = "sqlite:///databases/apdb.db"
+   config.isolation_level = "READ_UNCOMMITTED"
 
 .. prompt:: bash
 
-   make_apdb.py -C myApPipeConfig.py
+   make_apdb.py -C myApdbConfig.py
    ap_pipe.py repo --calib repo/calibs --rerun myrun -C myApPipeConfig.py --id [optional ID to process]
 
 .. _section-ap-pipe-apdb-seealso:
--- a/python/lsst/ap/pipe/make_apdb.py
+++ b/python/lsst/ap/pipe/make_apdb.py
@@ -80,7 +80,14 @@ The config overrides must define ``db_url`` to create a valid config.
         namespace = super().parse_args(args, namespace)
         del namespace.configfile
         namespace.config = ApdbConfig()
-        namespace.overrides.applyTo(namespace.config)
+        try:
+            namespace.overrides.applyTo(namespace.config)
+        except Exception as e:  # yes, configs really can raise anything
+            message = str(e)
+            if "diaPipe" in message:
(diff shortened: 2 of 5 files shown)


python/lsst/ap/pipe/make_apdb.py
```diff
@@ -73,7 +73,8 @@ def parse_args(self, args=None, namespace=None):
         """
         if not namespace:
             namespace = argparse.Namespace()
-        namespace.config = ApdbConfig()
+        # Make ApdbConfig as a subconfig of DiaPipelineConfig to ensure correct defaults get set
+        namespace.config = DiaPipelineConfig().apdb.value
```
mrawls, line 77: Can you please clarify for me what this is doing? I got the impression that `ApdbConfig`, `ApPipeConfig`, and `DiaPipelineConfig` are all entirely separate now, but this seems to imply one depends on another.

python/lsst/ap/pipe/make_apdb.py
```diff
@@ -108,8 +106,110 @@ def makeApdb(args=None):
...
-                         DiaSource=make_dia_source_schema()))
+    apdb = Apdb(config=parsedCmd.config,
+                afw_schemas=dict(DiaObject=make_dia_object_schema(),
+                                 DiaSource=make_dia_source_schema()))
     apdb.makeSchema()
     return apdb
+
+
+# --------------------------------------------------------------------
+# Copied from pipe_base to let us keep using them after Gen 2 is gone
```
mrawls, line 117: Thank you for this comment! Are there no other instances where we need an argument parser like this in gen3? I think putting this here is OK for the purposes of this ticket, but surely other pipetasks must have scripts one runs beforehand... ingestion, perhaps? Whatever it is we do now to declare a directory is a repo? Are other examples such as these just not as inherently configurable?

python/lsst/ap/pipe/make_apdb.py
```diff
@@ -108,8 +106,110 @@ def makeApdb(args=None):
...
+            Option value specified by the user.
+        """
+        if namespace.config is None:
+            return
+        for nameValue in values:
+            name, sep, valueStr = nameValue.partition("=")
+            if not valueStr:
+                parser.error(f"{option_string} value {nameValue} must be in form name=value")
+            if "diaPipe.apdb" in name:
+                parser.error(f"{nameValue} looks like it uses ApPipeConfig; try dropping 'diaPipe.apdb'")
```
mrawls, line 146: Thank you!

mrawls (commented):
I think this solves the issue at hand, but I do have a few lingering concerns about how we ensure users wind up with self-consistent configs when making an APDB vs. running the AP pipeline. Thank you for updating the docs and adding tests!

doc/lsst.ap.pipe/apdb.rst
```diff
@@ -28,14 +28,13 @@ Configuring the database
 ========================
 
 The database is configured using `~lsst.dax.apdb.ApdbConfig`.
-|ap_pipe| command line users can pass configuration information to the script through the :option:`--config <ap_pipe.py --config>` and :option:`--configfile <ap_pipe.py --configfile>` command-line options.
-|make_apdb| also uses `ApPipeConfig` and the :option:`--config` and :option:`--configfile` options, so users can pass exactly the same arguments to |make_apdb| and |ap_pipe|.
-Supporting identical command line arguments for both scripts makes it easy to keep the database settings in sync.
+|ap_pipe| command line users can pass configuration information to the script through the :option:`--config <ap_pipe.py --config>` and :option:`--configfile <ap_pipe.py --configfile>` command-line options, using the prefix ``diaPipe.apdb.`` to distinguish APDB information from other pipeline configuration.
```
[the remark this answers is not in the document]
kfindeisen: I think it would be pretty easy to add `--config-file` as an alias to `CmdLineTask`, and I _think_ we can avoid an RFC on the "nobody would seriously object" grounds. Though I might want to ping people on `#dm` just to avoid accusations of messing with shared code...

The `.` vs `:` distinction is more fundamental to both the Gen 2 and Gen 3 parsers, and I think we just have to live with it. Sorry!

I think both errors are fairly fail-fast, though. If you switch `--config-file` and `--configfile` you'll get a "No such option" error. If you don't use `:` in Gen 3 get "unrecognized option syntax". Using `:` in Gen 2 would probably get you a "config does not have a `diaPipe:apdb` field". Hopefully that's enough if we warn users about the differences.

doc/lsst.ap.pipe/apdb.rst
```diff
@@ -46,23 +45,26 @@ Databases can be configured using direct config overrides (see :ref:`ap-pipe-pip
 
 .. prompt:: bash
 
-   make_apdb.py -c diaPipe.apdb.isolation_level=READ_UNCOMMITTED diaPipe.apdb.db_url="sqlite:///databases/apdb.db" differencer.coaddName=dcr
+   make_apdb.py -c isolation_level=READ_UNCOMMITTED db_url="sqlite:///databases/apdb.db"
    ap_pipe.py -c diaPipe.apdb.isolation_level=READ_UNCOMMITTED diaPipe.apdb.db_url="sqlite:///databases/apdb.db" differencer.coaddName=dcr repo --calib repo/calibs --rerun myrun --id [optional IDs to process]
 
-|make_apdb| ignores any `ApPipeConfig` fields not related to the APDB (in the example, ``differencer.coaddName``), so there is no need to filter them out.
+The user is responsible for making sure the two APDB configurations match.
```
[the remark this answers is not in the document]
kfindeisen: There's a specific validation check in `ApdbConfig` that says "If you're using SQLite, you must use `READ_UNCOMMITTED`." The error could be clearer (doesn't say what field to set), but at least you know the config is wrong.

I'll switch the wording to "consistent".

doc/lsst.ap.pipe/apdb.rst
```diff
@@ -46,23 +45,26 @@ Databases can be configured using direct config overrides (see :ref:`ap-pipe-pip
...
 
 In Gen 3, this becomes (see :ref:`ap-pipe-pipeline-tutorial` for an explanation of |pipetask|):
 
 .. prompt:: bash
 
-   make_apdb.py -c diaPipe.apdb.isolation_level=READ_UNCOMMITTED diaPipe.apdb.db_url="sqlite:///databases/apdb.db" differencer.coaddName=dcr
+   make_apdb.py -c isolation_level=READ_UNCOMMITTED db_url="sqlite:///databases/apdb.db"
    pipetask run -p ApPipe.yaml -c diaPipe:apdb.isolation_level=READ_UNCOMMITTED diaPipe:apdb.db_url="sqlite:///databases/apdb.db" differencer:coaddName=dcr -b repo -o myrun
 
+Again, make sure the APDB is created and accessed with the same configuration; note that the pipeline file given by ``-p`` may include APDB config overrides.
```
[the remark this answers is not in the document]
kfindeisen: You mean a warning box for both "use the same config" and "pipelines may change the config"?

doc/lsst.ap.pipe/apdb.rst
```diff
@@ -46,23 +45,26 @@ Databases can be configured using direct config overrides (see :ref:`ap-pipe-pip
...
+You can double-check what configuration is being run by calling :command:`pipetask run` with the ``--show config="apdb*"`` argument, though this lists *all* configuration options, including those left at their defaults.
+
 Databases can also be set up using config files:
 
 .. prompt:: bash
 
-   make_apdb.py -C myApPipeConfig.py
+   make_apdb.py -C myApdbConfig.py
```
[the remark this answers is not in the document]
kfindeisen: It would be the same as the inline examples above, but with an extra `config.` in front. Could we just point people to the general config file documentation?

python/lsst/ap/pipe/make_apdb.py
kfindeisen: `ApdbConfig` can stand alone, but its default config isn't what AP actually uses. I needed to make sure [these defaults](https://github.com/lsst/ap_association/blob/master/python/lsst/ap/association/diaPipe.py#L145-L150) got called, because otherwise you get nasty errors about missing columns. This code is a bit of a hack, I just couldn't think of another way of doing things short of telling the user to configure `DiaPipelineConfig` (which would be the worst of both worlds, IMO).

python/lsst/ap/pipe/make_apdb.py
kfindeisen: Oops, you're right; the ingestion code uses [a standalone system](https://github.com/lsst/pipe_base/blob/master/python/lsst/pipe/base/configOverrides.py), so half the premise of this ticket is unnecessary. So now I have to rewrite everything. 😭

mrawls: Oh no, ingestion to the... rescue? Something like that. I'll be around whenever you're ready for review take 2. But it will be nice to avoid a surprising bit of argument parsing code living here.

doc/lsst.ap.pipe/apdb.rst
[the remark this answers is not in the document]
mrawls: I'm inclined to have a quote-text-box (essentially `cat myApdbConfig.py`) here *and also* link to the config documentation, if it's not too much trouble. `make_apdb.py` is an exception in so many other ways that users might not assume it uses any kind of standard configs either, and a quick example could enhance clarity.

doc/lsst.ap.pipe/apdb.rst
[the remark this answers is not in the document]
mrawls: Ooh I only meant the former, but the implications of the latter hadn't occurred to me, even though you did technically say it so... yes, both.

doc/lsst.ap.pipe/apdb.rst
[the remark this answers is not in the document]
mrawls: Fair, I'm OK overlooking this concern here since it's a bit bigger than this ticket. I'd love a lower-level equivalent to what you implemented for `diaPipe.apdb.stuff` returning "It looks like you typed `diaPipe.apdb.stuff`, maybe try `stuff`" for the `configfile` vs. `config-file` situation. I don't quite know how to write that ticket, but sounds like you've started the conversation on #dm-middleware - thanks.

python/lsst/ap/pipe/make_apdb.py
kfindeisen: Looks like I won't be able to completely avoid having some argument parsing code. The ingestion UI uses [Click](https://github.com/lsst/obs_base/blob/master/python/lsst/obs/base/cli/cmd/commands.py#L84-L101), and I really don't want to figure out how to hook that up to an independent script. 😬 

That said, using the new `ConfigOverrides` class makes the code that needs to live here a _lot_ shorter, so it's still a win.

python/lsst/ap/pipe/make_apdb.py
```diff
@@ -108,8 +116,43 @@ def makeApdb(args=None):
...
+                afw_schemas=dict(DiaObject=make_dia_object_schema(),
+                                 DiaSource=make_dia_source_schema()))
     apdb.makeSchema()
     return apdb
+
+
+# --------------------------------------------------------------------
+# argparse.Actions for use with ConfigOverrides
+# ConfigOverrides is normally used with Click, but I don't have time
+# to learn it.
```
mrawls, line 129: This comment is perhaps a bit too honest... I think it's OK to say ingest uses Click for this, but you chose argparse. The context here is that gen3 no longer has a config override mechanism but we need this functionality for make_apdb, right? You could explicitly say that here, but I suppose having it in the commit message is sufficient.

mrawls (approved):
Thanks for simplifying the config override situation!

PR #63, opened 2020-11-17
DM-26592: Create “afterburner” to analyse the results of AP processing with fakes included

morriscb:


kgabor:
Please combine the commits into one - I don't think that there are logical blocks here that need to be kept in separate commits.

kgabor (approved):

python/lsst/ap/pipe/matchApFakes.py
```diff
@@ -156,7 +176,19 @@ def trim(row):
         return fakeCat[fakeCat.apply(trim, axis=1)]
```
kgabor, line 177: I think adding "@staticmethod" here would make it clearer that this is a standalone internal calculation.

kgabor (approved):

python/lsst/ap/pipe/matchApFakes.py
morriscb: Eh, I'm fine with it the way it is. Not sure we use similar decorators much anywhere.

morriscb: Added a leading underscore to their names.

PR #65, opened 2020-12-08
DM-27388: Implement metric system for fakes in AP

morriscb:


python/lsst/ap/pipe/metrics.py
```diff
@@ -0,0 +1,138 @@
...
+from lsst.pipe.base import Struct
+import lsst.pipe.base.connectionTypes as connTypes
+from lsst.pipe.tasks.insertFakes import InsertFakesConfig
+from lsst.verify import Measurement, Name
+from lsst.verify.tasks import MetricTask, MetricComputationError
+
+
+class ApFakesCompletenessMetricConnections(
+        MetricTask.ConfigClass.ConnectionsClass,
+        dimensions={"instrument", "visit", "detector", "band"},
```
kfindeisen, line 43: I think the `band` is redundant here (it's implied by `visit`). Do you need to declare it to get the data ID extraction in `runQuantum` to work?

python/lsst/ap/pipe/metrics.py
```diff
@@ -0,0 +1,138 @@
...
+    """
+    _DefaultName = "apFakesCompleteness"
+    ConfigClass = ApFakesCompletenessMetricConfig
+
+    def runQuantum(self, butlerQC, inputRefs, outputRefs):
+        inputs = butlerQC.get(inputRefs)
+        inputs["band"] = butlerQC.quantum.dataId["band"]
+
+        outputs = self.run(**inputs)
+        butlerQC.put(outputs, outputRefs)
```
kfindeisen, line 102: Unfortunately, [`MetricTask.runQuantum`](https://github.com/lsst/verify/blob/master/python/lsst/verify/tasks/metricTask.py#L177-L189) has some error-handling logic that you've bypassed here; please add it in. Sorry about that; hopefully the extra code won't be necessary in the future.

python/lsst/ap/pipe/metrics.py
```diff
@@ -0,0 +1,138 @@
...
+        -------
+        result : `lsst.pipe.base.Struct`
+            A `~lsst.pipe.base.Struct` containing the following component:
+            ``measurement``
+                the ratio (`lsst.verify.Measurement` or `None`)
+        """
+        if matchedFakes is not None:
+            metricName = \
+                f"{self.config.metricName}"
+            magnitudes = matchedFakes[f"{self.config.magVar}" % band]
```
kfindeisen, line 124:
```suggestion
            metricName = self.config.metricName
            magnitudes = matchedFakes[self.config.magVar % band]
```

tests/test_metrics.py
```diff
@@ -0,0 +1,132 @@
...
+
+        fakesConfig = CreateRandomApFakesConfig()
+        fakesConfig.fraction = 0.0
+        fakesConfig.fakeDensity = self.sourceDensity
+        fakesTask = CreateRandomApFakesTask(config=fakesConfig)
+        fakeCat = fakesTask.run(self.tractId, self.simpleMap).fakeCat
+
+        self.band = 'g'
+        magCut = 25
+        magMask = (fakeCat[f"{fakesConfig.magVar}" % self.band] < magCut)
```
kfindeisen, line 78:
```suggestion
        magMask = (fakeCat[fakesConfig.magVar % self.band] < magCut)
```

tests/test_metrics.py
```diff
@@ -0,0 +1,132 @@
...
+        self.expectedAllMatched = magMask.sum()
+        ids = np.where(magMask, np.arange(1, len(fakeCat) + 1, dtype=int), 0)
+        self.fakeCat = fakeCat.assign(diaObjectId=ids,
+                                      filterName=["g"] * len(fakeCat),
+                                      diaSourceId=ids)
```
kfindeisen, line 83: This is a bit surprising. Shouldn't the catalog returned by `fakesTask` already have a filter column?

tests/test_metrics.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def testValid(self):
+        """Test the run method.
+        """
+        metricComplete = self.makeTask(20, 30)
```
kfindeisen, line 88: Isn't this equivalent to `self.task`?

tests/test_metrics.py
```diff
@@ -0,0 +1,132 @@
...
+        """
+        result = self.task.run(None, None)
+        lsst.pipe.base.testUtils.assertValidOutput(self.task, result)
+        meas = result.measurement
+        self.assertIsNone(meas)
+
+    def testValidEmpty(self):
+        """Test the run method with a valid but zero result.
+        """
+        metricComplete = self.makeTask(25, 30)
```
kfindeisen, line 109: I suggest making this
```suggestion
        metricComplete = self.makeTask(self.magCut, self.magCut + 5)
```
(with appropriate changes to `setUp()`). That way the test will be valid if the setup changes later.

kfindeisen (commented):
As brought up on Slack, there's not a good solution for what you're trying to do (and it's a much bigger question than a ticket about one metric). My comments below are based on the assumption that you take the AM1, AM2, etc. approach for now; that will at least keep measurements from different magnitude bins from overwriting each other.

python/lsst/ap/pipe/metrics.py
```diff
@@ -0,0 +1,138 @@
...
+    @property
+    def metricName(self):
+        """The metric calculated by a `MetricTask` with this config
+        (`lsst.verify.Name`, read-only).
+        """
+        return Name(package=self.connections.package,
+                    metric=f"{self.connections.metric}Mag{self.magMin}t{self.magMax}")
+
+
+class ApFakesCompletenessMetricTask(MetricTask):
```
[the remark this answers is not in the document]
morriscb: Yes. The code that creates the data for this metric is Gen3 only. I do not plan on making this backwards compatible.

tests/test_metrics.py
morriscb: The columns here are to mimic the results of matchFakes, not create fakes. The fakes have all magnitudes in all bands created. This is faking columns after processing.

python/lsst/ap/pipe/metrics.py
```diff
@@ -0,0 +1,138 @@
...
+from lsst.verify.tasks import MetricTask, MetricComputationError
+
+
+class ApFakesCompletenessMetricConnections(
+        MetricTask.ConfigClass.ConnectionsClass,
+        dimensions={"instrument", "visit", "detector", "band"},
+        defaultTemplates={"coaddName": "deep",
+                          "fakesType": "",
+                          "package": "ap_pipe",
+                          "metric": "apFakesCompleteness"}):
```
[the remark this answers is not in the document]
morriscb: I'm not quite getting where to set these names (This is in addition to your comment later about connections). Should I just remove the automated naming in this code and just rely on doing everything in the pipeline config? If so, what all do I have to set?

python/lsst/ap/pipe/metrics.py
```diff
@@ -0,0 +1,138 @@
...
+
+class ApFakesCompletenessMetricConfig(
+        MetricTask.ConfigClass,
+        InsertFakesConfig,
+        pipelineConnections=ApFakesCompletenessMetricConnections):
+    """ApFakesCompleteness config.
+    """
+    magMin = pexConfig.RangeField(
+        doc="Minimum magnitude the mag distribution. All magnitudes requested "
+            "are set to the same value.",
```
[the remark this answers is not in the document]
morriscb: Fixed

python/lsst/ap/pipe/metrics.py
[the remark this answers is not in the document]
kfindeisen: I meant here that you'd have to add the metrics (the final names, with magnitudes) to https://github.com/lsst/verify_metrics/blob/master/metrics/ap_pipe.yaml if you want them to work everywhere (I think SQuaSH only uploads metrics defined in those files). You can use `<<`  notation (see [pipe_tasks.yaml](https://github.com/lsst/verify_metrics/blob/master/metrics/pipe_tasks.yaml) for examples) to cut down on the boilerplate.

Doing everything in the pipeline config would work; I think you'd just have to set `connections.metric` plus the magnitude values for each case.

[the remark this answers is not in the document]
kfindeisen: So something like:
```
apFakesCompleteness20t25: &apFakesCompleteness
    description: ...
    unit: ""
    reference:
        url: https://confluence.lsstcorp.org/pages/viewpage.action?spaceKey=~ebellm&title=Alert+Production+Metrics
    tags:
        - ap_verify
        - assoc
        - CCD

apFakesCompleteness25t30:
    <<: *apFakesCompleteness
    description: ...
```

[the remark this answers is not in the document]
morriscb: Okay, I basically have another ticket for this set of work: DM-26593. This one is mostly just to make the metric and get it tested so I'll add adding this to that ticket.

PR #70, opened 2021-02-04
DM-28092: Document that DiaPipeTask can only handle specific bands

morriscb:


parejkoj (commented):
Some wording suggestions

parejkoj (approved):

PR #75, opened 2021-04-20
DM-29147: Switch from using mapApData to new pipeline task and functors.

morriscb:
Debug ApPipe use of transformDiaSourceCat.


tests/test_appipe.py
```diff
@@ -111,9 +111,11 @@ def mockPatchSubtasks(self, task):
         """
         with patch.object(task, "ccdProcessor") as mockCcdProcessor, \
                 patch.object(task, "differencer") as mockDifferencer, \
+                patch.object(task, "transformDiaSrcCat") as mockTransForm, \
```
parejkoj, line 114: This is strange capitalization. Shouldn't it be `mockTransform` (and below)?

parejkoj (commented):
Please fix the `Dm` in the PR title to `DM`.

Some wording comments.

PR #76, opened 2021-05-04
DM-30015: Update doc/conf.py for documenteer 0.6

jonathansick:
Updates based on the latest updates in the package template (https://github.com/lsst/templates/tree/master/project_templates/stack_package)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -5,7 +5,7 @@ ignore = E133, E226, E228, N802, N803, N806, N812, N813, N815, N816, W503, E266
 exclude =
   bin,
   config,
-  doc,
+  doc/conf.py,
   **/*/__init__.py,
   **/*/version.py,
   tests/.tests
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -5,7 +5,7 @@ ignore = E133, E226, E228, N802, N803, N806, N812, N813, N815, N816, W503, E266
 exclude =
   bin,
   config,
-  doc,
+  doc/conf.py,
```
kfindeisen, line 8: Is this change correct? I tried converting the config in `pipe_tasks`, and I got errors from `doc/_build/html/conf.py` unless I excluded all of `doc`.

jonathansick: huh, and I got reports from other devs that conf.py wasn't being ignored if I only listed `doc`! I'll just update `documenteer` to ensure that `conf.py` isn't copied into the build (which I what I should have been doing all along). Thanks for letter me know.

kfindeisen: It might depend on the flake8 version? I used to get errors from `doc/conf.py` with the old ignore, but now I don't. 😕

PR #79, opened 2021-06-01
remove some imageDifference options that are now in task defaults

kherner:


python/lsst/ap/pipe/ap_pipe.py
```diff
@@ -62,12 +62,6 @@ class ApPipeConfig(pexConfig.Config):
     def setDefaults(self):
         """Settings appropriate for most or all ap_pipe runs.
         """
-        # Always prefer decorrelation; may eventually become ImageDifferenceTask default
-        self.differencer.doDecorrelation = True
-        self.differencer.detection.thresholdValue = 5.0  # needed with doDecorrelation
-
-        # Don't have source catalogs for templates
-        self.differencer.doSelectSources = False
```
kfindeisen, line 70: I think that, in the case of `ap_verify`, this actually belongs in our dataset configs (since whether or not we have template source catalogs is a function of how the templates were generated). @mrawls, is this reasoning correct? I noticed this got changed in the task defaults, so I'm wondering if there's a more fundamental reason for `doSelectSources = False`.

mrawls: I am not immediately aware of a situation where we want to use template source catalogs in image differencing, even if they do happen to exist. I'm sure we could come up with a situation (DRP?), but changing the task default to line up with what we almost always use seems correct to me.

I like the idea of configuring this at the `ap_verify` dataset level when the new `doSelectSources = False` default is not appropriate.

PR #80, opened 2021-06-04
DM-29338: Incorporate Gen 3 crosstalk in HiTS runs

mrawls:
This PR updates a whole bunch of pipelines and adopts a new default AP Pipeline template name, `goodSeeing` (instead of `deep`).


bps/bps_ApTemplate.yaml
```diff
@@ -0,0 +1,47 @@
...
+# Format for job names and job output filenames
+templateDataId: '{tract}_{patch}_{band}_{visit}_{exposure}_{detector}'
+
+# Just names, but UPDATE THIS AND KEEP THEM SHORT
+project: ApTemplate
+campaign: DM-xxxxx
+
+submitPath: ${PWD}/bps/{outCollection}
+computeSite: ncsapool
+# Memory allocated for each quantum, in MBs
```
isullivan, line 13: Do you know the maximum that can be requested? If so, that would be good to include.

bps/bps_ApTemplate.yaml
```diff
@@ -0,0 +1,47 @@
...
+  # UPDATE THIS to also set the output collection name
+  payloadName: DM-xxxxx
+  # UPDATE THIS to point to the correct repository
+  butlerConfig: /repo/dc2/butler.yaml
+  # UPDATE THIS and be sure it includes collections with raws, calibs, refcats, and skymaps
+  inCollection: 2.2i/defaults/test-med-1
+  # Note: keep timestamp in outCollection so you don't get a zillion sub-runs
+  output : 'u/${USER}/{payloadName}'
+  outCollection: '{output}/{timestamp}'
+  # UPDATE THIS to specify what data to process
```
isullivan, line 30: Perhaps expand this comment to include the additional keys beyond "band" and "instrument" that people might want to specify.

isullivan (approved):
Very helpful comments and descriptions throughout, thank you.

bps/bps_ApTemplate.yaml
mrawls: This is a good suggestion, but I have empirically found that doing DECam runs with complex data queries can lead to an extremely lengthy (many days) query time. The example is intentionally minimal, but may be able to be a bit more instructive in the future when this is resolved.

bps/bps_ApTemplate.yaml
mrawls: I don't know, and therefore haven't included it 🙃 in practice, 2000-4000 is enough for most jobs, and bumping it up to 12000 or so is enough to get the occasional held job through.

PR #81, opened 2021-06-08
DM-30563: ap_verify failing to find dataset type fakes_deepDiff_warpedExp

kfindeisen:
This PR documents the configs in `ApPipe.yaml` and extends them to cover all coadd-related connections in the main pipeline chain. This prevents downstream pipeline edits that only touch templates from creating inconsistent configs.

### DIFF
--- a/pipelines/ApPipe.yaml
+++ b/pipelines/ApPipe.yaml
@@ -23,9 +23,15 @@ tasks:
       getTemplate.coaddName: goodSeeing  # Can be removed once ImageDifference no longer supports Gen 2
       connections.coaddName: goodSeeing
       # TODO: redundant connection definitions workaround for DM-30210
+      connections.exposure: calexp
       connections.coaddExposures: goodSeeingCoadd
+      connections.dcrCoadds: dcrCoadd
+      connections.outputSchema: goodSeeingDiff_diaSrc_schema
       connections.subtractedExposure: goodSeeingDiff_differenceExp
+      connections.scoreExposure: goodSeeingDiff_scoreExp
       connections.warpedExposure: goodSeeingDiff_warpedExp
+      connections.matchedExposure: goodSeeingDiff_matchedExp
+      connections.diaSources: goodSeeingDiff_diaSrc
       # TODO: end DM-30210 workaround
   transformDiaSrcCat:
     class: lsst.ap.association.TransformDiaSourceCatalogTask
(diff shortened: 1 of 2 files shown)


pipelines/ApPipe.yaml
```diff
@@ -17,17 +17,26 @@ tasks:
   imageDifference:
     class: lsst.pipe.tasks.imageDifference.ImageDifferenceTask
     config:
-      doWriteWarpedExp: True
-      coaddName: goodSeeing
-      getTemplate.coaddName: goodSeeing
+      doWriteWarpedExp: True             # Required for diaPipe
+      # 'goodSeeing' expected type for LSST processing; task default 'deep' for Gen 2 pipelines
+      coaddName: goodSeeing              # Redundant line for Gen 2 compatibility
+      getTemplate.coaddName: goodSeeing  # Redundant line for Gen 2 compatibility
```
parejkoj, line 23: I'm confused about the "gen2 compatibility" statement here: this file is only used in gen3. The comment above says that "goodSeeing" is the correct name here, but the "redundant" line suggests these are unnecessary?

pipelines/ApPipe.yaml
```diff
@@ -17,17 +17,26 @@ tasks:
...
       connections.coaddName: goodSeeing
+      # TODO: redundant connection definitions workaround for DM-30210
```
parejkoj, line 25: Maybe put another mark at the bottom of the redundant block? (similarly below)

parejkoj (commented):
I'm confused about the meaning of some of the comments here.

pipelines/ApPipe.yaml
kfindeisen: `coaddName` and `getTemplate.coaddName` are used by Gen 2, and the config requires that they and `connections.coaddName` all be consistent with each other. "Redundant" in the sense that ideally the coadd name would only need to be specified in one place.

parejkoj (approved):

pipelines/ApPipe.yaml
```diff
@@ -17,31 +17,47 @@ tasks:
...
+      connections.diffIm: goodSeeingDiff_differenceExp
+      connections.warpedExposure: goodSeeingDiff_warpedExp
       connections.associatedDiaSources: goodSeeingDiff_assocDiaSrcTable
+      # TODO: end DM-30210 workaround
 contracts:
-  # DiaPipelineTask needs diaSource fluxes, catalogs, and difference exposures
+  # DiaPipelineTask needs diaSource fluxes, catalogs, warped exposures, and difference exposures
   - imageDifference.doMeasurement is True
   - imageDifference.doWriteSources is True
+  - imageDifference.doWriteWarpedExp is True
```
kfindeisen, line 60: Just realized we should also add this check to the Gen 2 pipeline...

PR #85, opened 2021-07-19
DM-31064: Create metric to capture the total number of fake sources used

morriscb:


python/lsst/ap/pipe/metrics.py
```diff
@@ -139,3 +140,57 @@ def run(self, matchedFakes, band):
...
+    _DefaultName = "apFakesCount"
+    ConfigClass = ApFakesCountMetricConfig
+
+    def run(self, matchedFakes, band):
+        """Compute the completeness of recovered fakes within a magnitude
+        range.
+
+        Parameters
+        ----------
+        matchedFakes : `lsst.afw.table.SourceCatalog` or `None`
```
isullivan, line 176: Please add a Parameters docstring for `band`

python/lsst/ap/pipe/metrics.py
```diff
@@ -139,3 +140,57 @@ def run(self, matchedFakes, band):
...
+                the ratio (`lsst.verify.Measurement` or `None`)
+        """
+        if matchedFakes is not None:
+            magnitudes = matchedFakes[f"{self.config.magVar}" % band]
+            magCutFakes = matchedFakes[np.logical_and(magnitudes > self.config.magMin,
+                                                      magnitudes < self.config.magMax)]
+            meas = Measurement(self.config.metricName,
+                               len(magCutFakes) * u.count)
+        else:
+            self.log.info("Nothing to do: no matched catalog found.")
```
isullivan, line 194: Do you expect this task to be run on data without a matched catalog? If it is not expected that `matchedFakes` would be `None`, then this should at least be `self.log.warn`.

isullivan: I might suggest slightly changing the wording, as well, to reflect that this task does not try to load a catalog itself: "Nothing to do: no matched catalog supplied."

isullivan (approved):
Looks good; a few documentation suggestions.

python/lsst/ap/pipe/metrics.py
morriscb: I'm okay with this being info. It's consistent with the other metrics we current have.

PR #89, opened 2021-09-08
DM-26140: Standardize AP pipeline structure

kfindeisen:
This PR reorganizes the `ap_pipe` pipeline structure slightly to make it easier to call from `ap_verify`. The only functional change should be the creation of new subsets.


mrawls (commented):
Thank you for greatly improving these pipelines across so many cameras! Please ensure they all use two-space indents; I think there are a few files with four space indents.

kfindeisen:
I thought I managed to be consistent within a single file. If I need to standardize the files, then why two spaces instead of four? I'd find two spaces hard to read, especially for long blocks.

mrawls:
We still need a "real" yaml style guide, don't we... I'm basing my interim 2-space assertion on what `DRP.yaml` uses and what @jonathansick said on Slack https://lsstc.slack.com/archives/C2JPL2DGD/p1628187106046800.

PR #95, opened 2021-11-03
DM-32430: Create DECam DRP Processing Pipeline

leeskelvin:


kfindeisen:
Sorry, but why on earth does `ap_pipe` need a DRP pipeline?

leeskelvin:
Ah, now it doesn't. This PR was created before `drp_pipe` came into existence, as it was agreed that all DECam-related pipelines were to live temporarily in `ap_pipe` (rather than `obs_decam`). This PR will be closed shortly.

PR #98, opened 2022-01-21
PREOPS-931: make multi-tract image differencing default in ApPipe.yaml

kherner:


kfindeisen (approved):
Looks good; only minor comments for the pipeline itself. However, please clean up the commits as recommended by the [DM style guide](https://developer.lsst.io/work/flow.html#git-commit-organization-best-practices). In particular:

- three of the messages have "make multi-tract image differencing default in ApPipe.yaml" as their brief text, making it hard to tell what they actually do and what the difference is.
- the second commit is worded as if it were a personal note; I suggest either rephrasing to something like "Add retrieveTemplate to apPipe subset" or merging with the first commit (on the grounds that, without the subset update, the pipeline is not self-consistent).
- the last two commits together are effectively a reformat; I suggest either deleting both, or merging them together into a "Rearrange ApPipe contracts" commit (with an explanation of why the new order is preferable?).

pipelines/ApPipe.yaml
```diff
@@ -24,10 +24,16 @@ parameters:
   diffWarp: goodSeeingDiff_warpedExp
   diffMatch: goodSeeingDiff_matchedExp
   assocSrc: goodSeeingDiff_assocDiaSrc
+  templateExp: goodSeeingDiff_templateExp
   # TODO: end DM-30210 workaround
 tasks:
+  retrieveTemplate:  # For multi-tract difference imaging
+    class: lsst.ip.diffim.getTemplate.GetMultiTractCoaddTemplateTask
+    config:
+      connections.coaddExposures: goodSeeingCoadd
```
kfindeisen, line 33: Didn't notice this before:
```suggestion
      connections.coaddExposures: parameters.template
```
(Perhaps the parameter should be renamed, but it might be better to do that on a different ticket.)

mrawls:
Ken and I moved these changes to DM-32988 (#99) to keep all code changes on DM ticket branches per current policy.

PR #104, opened 2022-04-21
DM-33692: HSC's ApTemplate pipeline is missing skyCorr step

mrawls:
This PR adds the skyCorrectionTask step to ApTemplate.yaml for HSC. I also fixed a typo.


kmrakovcic (approved):
Makes sense!

PR #106, opened 2022-05-04
DM-32694: Split AP pipeline into ApPipeWithFakes

kfindeisen:
This PR significantly reworks the AP pipelines, making two major changes:

- the [DM-30210](https://jira.lsstcorp.org/browse/DM-30210) workarounds are now done at the contract level, not the parameter/config level. This arrangement is harder to read, but is much more portable to importing pipelines.
- the `ApPipeWithFakes` pipeline no longer imports `ApPipe`, giving it independent task labels and contracts. This change is strictly for the benefit of `ap_verify` (which needs to be careful about distinguishing fakes and non-fakes tasks), and does not improve `ap_pipe` itself.

### DIFF
--- a/pipelines/ApPipe.yaml
+++ b/pipelines/ApPipe.yaml
@@ -14,25 +14,11 @@ imports:
 parameters:
   # Pipeline configurable to run on both goodSeeing and deep templates, depending on dataset.
   coaddName: goodSeeing
-  # TODO: redundant connection definitions workaround for DM-30210
-  template: goodSeeingCoadd
-  diaSrcCat: goodSeeingDiff_diaSrc
-  diaSrcSchema: goodSeeingDiff_diaSrc_schema
-  diaSrcParquet: goodSeeingDiff_diaSrcTable
-  diff: goodSeeingDiff_differenceExp
-  diffScore: goodSeeingDiff_scoreExp
-  diffWarp: goodSeeingDiff_warpedExp
-  diffMatch: goodSeeingDiff_matchedExp
-  assocSrc: goodSeeingDiff_assocDiaSrc
-  templateExp: goodSeeingDiff_templateExp
-  # TODO: end DM-30210 workaround
 tasks:
   retrieveTemplate:  # For multi-tract difference imaging
     class: lsst.ip.diffim.getTemplate.GetTemplateTask
     config:
       connections.coaddName: parameters.coaddName
-      connections.coaddExposures: parameters.template
-      connections.outputExposure: parameters.templateExp
   imageDifference:
     class: lsst.pipe.tasks.imageDifference.ImageDifferenceFromTemplateTask
     config:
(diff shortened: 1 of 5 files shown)


pipelines/ApPipe.yaml
```diff
@@ -92,9 +55,18 @@ contracts:
   - imageDifference.doWriteSubtractedExp is True
   - imageDifference.doSkySources == transformDiaSrcCat.doRemoveSkySources
   # Inputs and outputs must match
-  - retrieveTemplate.connections.outputExposure == imageDifference.connections.inputTemplate
-  - imageDifference.connections.coaddName == transformDiaSrcCat.connections.coaddName
-  - imageDifference.connections.fakesType == transformDiaSrcCat.connections.fakesType
-  - imageDifference.connections.coaddName == diaPipe.connections.coaddName
-  - imageDifference.connections.fakesType == diaPipe.connections.fakesType
-  - transformDiaSrcCat.connections.diaSourceTable == diaPipe.connections.diaSourceTable
+  # Use of ConnectionsClass for templated fields is a workaround for DM-30210
```
mrawls, line 58: Yay much nicer to have the mess in one place here, as in ap_verify!

pipelines/ApPipeWithFakes.yaml
```diff
@@ -31,45 +25,109 @@ tasks:
...
       # apdb.db_url: YOU MUST CONFIGURE THIS BEFORE RUNNING THE PIPELINE
   fakesMatch:
-    class: lsst.pipe.tasks.matchFakes.MatchFakesTask
+    class: lsst.pipe.tasks.matchFakes.MatchVariableFakesTask
     config:
       connections.coaddName: parameters.coaddName
+      connections.fakesType: parameters.fakesType
       matchDistanceArcseconds: 0.5
+subsets:
+  apPipe:
```
mrawls, line 73: I think there needs to be a `processCcd` subset in here somewhere, because the camera-specific pipelines exclude it (in favor of camera-specific `processCcd`). Or maybe that is handled when `ProcessCcd.yaml` is imported up at the top? Mostly I am confused because `apPipe` as a subset looks like what my brain categorizes as just `processCcd`.

Basically, please convince me that the camera-specific `ApPipeWithFakes` pipelines are handling where they import "processCcd" steps and from correctly.

pipelines/HyperSuprimeCam/ApPipeWithFakes.yaml
```diff
@@ -21,18 +21,18 @@ imports:
...
-    # Do not integrate file and pipeline configs until obs config files are
-    # gone, to make it easier to check for changes on the obs side.
-    # Config file wipes out all pre-existing configs, so copy base pipeline
-    # config on top.
-    connections.coaddName: parameters.coaddName
-    insertFakes.doSubSelectSources: True
-    insertFakes.select_col: 'isVisitSource'
-    calibrate.photoCal.match.referenceSelection.magLimit.fluxField: i_flux
-    calibrate.photoCal.match.referenceSelection.magLimit.maximum: 22.0
+    class: lsst.pipe.tasks.processCcdWithFakes.ProcessCcdWithVariableFakesTask
```
mrawls, line 24: Aaahhhhh indentation fail, good catch

mrawls (commented):
If you can please convince me that each of the camera-specific `ApPipeWithFakes` pipelines are importing the correct `processCcd` steps (isr, calibrate, characterize), then I will be very thrilled with all of this 😀

pipelines/ApPipeWithFakes.yaml
kfindeisen: I think `processCcd` is unchanged from the import, since we don't exclude anything from `ProcessCcd.yaml`. I could add a comment here to that effect.

I'm not sure what to do about the `apPipe` subset. I originally wrote it intending to add a non-fakes `imageDifference` step, but later decided that would be silly (the only reason we need such a task is so that certain metrics won't be contaminated by fakes). I want the subsets to be consistent with what `ap_verify` is using, so maybe just `prepareFakes` and `apPipeWithFakes`?

kfindeisen: I'm not sure what to do about the `apPipe` subset. I originally wrote it intending to add a non-fakes `imageDifference` step to `ApPipeWithFakes`, but later decided that would be silly (the only reason we need such a task is so that certain `ap_verify` metrics won't be contaminated by fakes). I want the subsets to be consistent with what `ApVerifyWithFakes` is using, so maybe just `prepareFakes` and `apPipeWithFakes`?

kfindeisen:
I hope I answered your question about `processCCd`.

More broadly, are you OK with the fact that `ApPipeWithFakes` has been largely rewritten? While I'm not surprised you approve of the new contract handling, I expected a bit more resistance to the restructuring/renaming...

pipelines/ApPipeWithFakes.yaml
mrawls: I honestly never use the subsets, which is probably why there weren't any in here to begin with. In my mind, though, an `apPipe` subset ought to be defined as processCcd + imageDifference + diaPipe (where processCcd = isr + characterize + calibrate). Defining it as such - and calling it whatever you like, e.g., either `apPipe` or `apPipeWithFakes` - would work fine for me.

mrawls: And sure, a `prepareFakes` subset sounds useful!

mrawls:
I'm good with the restructuring you did to make the fakes pipeline stand alone - there are arguments to be made for having it import more from regular ApPipe, but I see why doing it this way makes ap_verify's life easier, and the labels are more clear and less likely to accidentally... I don't know... make a template with fakes and forget to use it 🙃 

My only quibbles are with definitions of the subsets, as discussed above. If we can sort that out, AND if you're sure ap_verify (which imports the new ApPipeWithFakes pipelines) runs fine on both DECam and HSC datasets, then I'm quite happy.

pipelines/ApPipeWithFakes.yaml
kfindeisen: Ok, I'll remove the `apPipe` subset then.

PR #107, opened 2022-05-05
DM-34254: Update piff psf kernelSize to 25

parejkoj:


ebellm (approved):

pipelines/ProcessCcd.yaml
```diff
@@ -1,7 +1,13 @@
...
 tasks:
   isr: lsst.ip.isr.IsrTask
-  characterizeImage: lsst.pipe.tasks.characterizeImage.CharacterizeImageTask
+  characterizeImage:
+    class: lsst.pipe.tasks.characterizeImage.CharacterizeImageTask
+    # We want a larger psf kernel, to include the calibFlux 12 pixel radius.
+    config:
+      python: |
+        config.measurePsf.makePsfCandidates.kernelSize = 25
+        config.measurePsf.psfDeterminer["piff"].kernelSize = 25
```
kfindeisen, line 10: I think you'd also need to add this to the instrument-specific pipelines after all, because they have a `characterizeImage.py` with the config-reset hack. 😞

kfindeisen: Hmm... no, it looks like it gets propagated. I think that means I was wrong about the evaluation order. 😕

kfindeisen: Jim Bosch was surprised when I asked about this behavior and suggested it could be a bug, so I don't think we can count on the current behavior in the future.

PR #108, opened 2022-05-06
DM-34699: Patch the ap_pipe config reset hack

kfindeisen:
This PR fixes a bug in the configs added in #102, and fixes some pipeline bugs that were discovered in the process of testing the fix.

### DIFF
--- a/pipelines/HyperSuprimeCam/ProcessCcd.yaml
+++ b/pipelines/HyperSuprimeCam/ProcessCcd.yaml
@@ -10,6 +10,11 @@ tasks:
       file: $AP_PIPE_DIR/config/HSC/characterizeImage.py
       # Do not integrate file and pipeline configs until obs config files are
       # gone, to make it easier to check for changes on the obs side.
+      # Config file wipes out all pre-existing configs, so copy base pipeline
+      # config on top.
+      python: |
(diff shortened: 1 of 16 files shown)


pipelines/HyperSuprimeCam/ProcessCcd.yaml
```diff
@@ -10,6 +10,11 @@ tasks:
       file: $AP_PIPE_DIR/config/HSC/characterizeImage.py
       # Do not integrate file and pipeline configs until obs config files are
       # gone, to make it easier to check for changes on the obs side.
+      # Config file wipes out all pre-existing configs, so copy base pipeline
+      # config on top.
```
parejkoj, line 14: Because of this comment, does it mean we have to copy all configs from the base processCcd.yaml? Or has that already been done? Could you work around it by doing a `file:ProcessCcd.yaml` above here?

parejkoj (commented):
How is this different from the other approach? How do you demonstrate that it does what we want?

pipelines/HyperSuprimeCam/ProcessCcd.yaml
kfindeisen: No, there's no such thing as importing a pipeline into a task config. There's a global pipeline import, but the `characterizeImage.py` gets applied after it.

PR #113, opened 2022-06-08
DM-34698: Remove config overrides setting kernelSize to 25

arunkannawadi:
These were explicitly set to 25 when drp_pipe wanted kernelSize to be 21
during their validation. Now that the default is switched to 25, these
overrides are no longer required.


parejkoj (approved):
Thanks for clearing this up.

PR #122, opened 2022-07-30
DM-34711: Reprocess HSC Cosmos dataset with Fakes

mrawls:
This PR fixes some HSC configs that caused an error when running ApPipeWithFakes. The task injects fakes to visits is sneaky because it calls calibrateTask a second time with most things (like astrometry and photometry) turned off. The PR also adds a config file for DECam so its visit-level fakes injection properly picks up camera-specific calibrateTask configs too.

### DIFF
--- a/config/DECam/processCcdWithFakes.py
+++ b/config/DECam/processCcdWithFakes.py
@@ -1,0 +1,34 @@
+# This file is modeled after HSC's, and has all of the minimal DECam-specific
+# calibrate configs. Since processCcdWithFakesTask uses a minimal version of
+# calibrateTask, we can't import the whole thing, because it configs for
+# astrometry and photometry; these are turned off in processCcdWithFakes.
+
+import os.path
+
+from lsst.meas.algorithms import ColorLimit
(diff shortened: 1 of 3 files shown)


config/DECam/processCcdWithFakes.py
```diff
@@ -0,0 +1,36 @@
...
+import os.path
+
+from lsst.meas.algorithms import ColorLimit
+from lsst.meas.astrom import MatchOptimisticBConfig
+
+# HACK: Throw away any changes imposed by obs configs, especially plugins.
+config.loadFromString(type(config)().saveToString())
+
+obsConfigDir = os.path.join(os.path.dirname(__file__))
+# calFile = os.path.join(ObsConfigDir, "calibrate.py")
```
kherner, line 15: Any reason not to delete these commented out lines if we don't need them?

kherner (approved):

config/DECam/processCcdWithFakes.py
mrawls: Good point, can delete

PR #124, opened 2022-08-20
DM-35817: Remove deprecated ref_dataset_name

parejkoj:


pipelines/DarkEnergyCamera/ApPipe_hits2015-3.yaml
```diff
@@ -6,13 +6,9 @@ imports:
 tasks:
   characterizeImage:
     class: lsst.pipe.tasks.characterizeImage.CharacterizeImageTask
-    config:
-      refObjLoader.ref_dataset_name: 'panstarrs'
   calibrate:
     class: lsst.pipe.tasks.calibrate.CalibrateTask
     config:
-      astromRefObjLoader.ref_dataset_name: 'gaia'
-      photoRefObjLoader.ref_dataset_name: 'panstarrs'
```
kfindeisen, line 15: If we're editing an archival file, should we just delete it (possibly question for @mrawls)?

mrawls: This whole file can definitely go away, since it refers to a repo that no longer exists

PR #127, opened 2023-01-10
DM-35207: Reset MakeWarp visitSummary connection in ApTemplate.yaml

TallJimbo:
Running updateVisitSummary here (as is done in DRP) would be wasteful, as there are no updated calibration objects being produced in this pipeline.

### DIFF
--- a/pipelines/ApTemplate.yaml
+++ b/pipelines/ApTemplate.yaml
@@ -21,6 +21,10 @@ tasks:
   makeWarp:
     class: lsst.pipe.tasks.makeCoaddTempExp.MakeWarpTask
     config:
+      # The upstream default is now finalVisitSummary, which doesn't exist
+      # in this pipeline since there's nothing to update relative to the
+      # original visitSummary.
+      connections.visitSummary: visitSummary
       doWriteEmptyWarps: True
       doApplyExternalPhotoCalib: False
       doApplyExternalSkyWcs: False
--- a/pipelines/ApTemplate.yaml
+++ b/pipelines/ApTemplate.yaml
@@ -19,7 +19,7 @@ tasks:
     config:
       connections.coaddName: parameters.coaddName
   makeWarp:
-    class: lsst.pipe.tasks.makeCoaddTempExp.MakeWarpTask
+    class: lsst.pipe.tasks.makeWarp.MakeWarpTask
     config:
       # The upstream default is now finalVisitSummary, which doesn't exist
       # in this pipeline since there's nothing to update relative to the


parejkoj:
Can you please put a comment line above your addition to explain why it's there? I was trying to figure it out from the PR description and commit description, but even those took me a bit. Someone looking at the pipeline will probably wonder why this isn't part of the default pipeline.

laurenam (approved):

pipelines/ApTemplate.yaml
```diff
@@ -21,6 +21,7 @@ tasks:
   makeWarp:
     class: lsst.pipe.tasks.makeCoaddTempExp.MakeWarpTask
```
kfindeisen, line 22:
```suggestion
    class: lsst.pipe.tasks.makeWarp.MakeWarpTask
```

pipelines/ApTemplate.yaml
```diff
@@ -21,6 +21,7 @@ tasks:
...
     config:
+      connections.visitSummary: visitSummary
```
kfindeisen, line 24: Since this is the template-generation pipeline, it's worth asking whether we actually _want_ to use `finalVisitSummary` here. The usual objection that final stuff doesn't exist in the AP pipeline doesn't apply. However, I assume you verified that the `consolidateVisitSummary.connections.ConnectionsClass(config=consolidateVisitSummary).visitSummary.name ==
      makeWarp.connections.ConnectionsClass(config=makeWarp).visitSummary.name` contract still passes.

kfindeisen (commented):
Just wanted to check that we really do want non-final summaries here, since this is not part of the AP pipeline proper.

pipelines/ApTemplate.yaml
TallJimbo: Considering the rest of this pipeline fixed, I don't think you want to add the `updateVisitSummary` task in this pipeline to create `finalVisitSummary`, because it wouldn't actually have any updates to apply relative to what's in `visitSummary`.  It might be worth considering renaming the output of `consolidateVisitSummary` task here to `finalVisitSummary` instead, so that PipelineTasks that consume it could stick to their (new) defaults.  

You'd also certainly want it if you ever wanted to start running jointcal, doing a second round of PSF modeling, or doing full focal plane background subtraction - but I think if that was ever the case, you'd really just want to start using a DRP pipeline, even if we'd need to reorganize the package dependency structure to make that viable.  (I assume that's what we'll do to make templates in production anyway, even in year 1).

I have run `ap_verify` on `ap_verify_ci_hits2015` locally, so I assume that will have checked the contracts.

PR #129, opened 2023-01-20
DM-37622: changes for HSMshape as the default

kherner:


pipelines/ApPipe.yaml
```diff
@@ -35,6 +35,11 @@ tasks:
     config:
       connections.coaddName: parameters.coaddName
       doSkySources: True
+      python: |
+        import lsst.meas.extensions.shapeHSM
+        config.measurement.slots.psfShape = "ext_shapeHSM_HsmPsfMoments"
+        config.measurement.slots.shape = "ext_shapeHSM_HsmSourceMoments"
+        config.measurement.plugins.names |= ["ext_shapeHSM_HsmSourceMoments", "ext_shapeHSM_HsmPsfMoments"]
```
parejkoj, line 42: Why can't we make these changes in `DetectAndMeasureConfig` directly? It's a diffim-specific task, and we probably want its default config to be appropriate for measuring on difference images.

pipelines/ApPipeWithFakes.yaml
```diff
@@ -59,6 +59,11 @@ tasks:
       connections.coaddName: parameters.coaddName
       connections.fakesType: parameters.fakesType
       doSkySources: True
+      python: |
+        import lsst.meas.extensions.shapeHSM
+        config.measurement.slots.psfShape = "ext_shapeHSM_HsmPsfMoments"
+        config.measurement.slots.shape = "ext_shapeHSM_HsmSourceMoments"
+        config.measurement.plugins.names |= ["ext_shapeHSM_HsmSourceMoments", "ext_shapeHSM_HsmPsfMoments"]
```
parejkoj, line 66: Same comment here: we should just set this in the Task Config `setDefaults`.

parejkoj (changes requested):

pipelines/ApPipe.yaml
kherner: Agreed; I've opened a PR for ip_diffim to do just that: https://github.com/lsst/ip_diffim/pull/247

pipelines/ApPipeWithFakes.yaml
kherner: Done in the ip_diffim PR. I'm probably going to just cancel/close this PR and delete the branch entirely.

kherner:
We're going to do the changes in the setDefaults function for DetectAndMeasureConfig in detectAndMeasure.py, so we don't need this branch or PR now.

PR #123, opened 2022-08-09
DM-35670: Update location of makeWarp task.

erykoff:

PR #134, opened 2023-03-10
DM-38321: Remove unused config item from AP HSC coaddBase

leeskelvin:


kfindeisen:
Closes #123.

config/HSC/makeWarp.py
```diff
@@ -15,9 +15,9 @@
 config.doApplySkyCorr = True
 
 config.modelPsf.defaultFwhm = 7.7
-config.warpAndPsfMatch.psfMatch.kernel['AL'].alardSigGauss = [1.0, 2.0, 4.5]
-config.warpAndPsfMatch.warp.warpingKernelName = 'lanczos5'
-config.coaddPsf.warpingKernelName = 'lanczos5'
+config.warpAndPsfMatch.psfMatch.kernel["AL"].alardSigGauss = [1.0, 2.0, 4.5]
+config.warpAndPsfMatch.warp.warpingKernelName = "lanczos5"
+config.coaddPsf.warpingKernelName = "lanczos5"
```
kfindeisen, line 20: In the future, please make such formatting changes on a separate commit. That would also help clarify why you made the change; AFAIK `"` and `'` are pretty much interchangeable in Science Pipelines.

kfindeisen (approved):
Changes look good, my only comment is procedural.

config/HSC/makeWarp.py
leeskelvin: Commits split. Formatting changes made in line with [current style guide formatting recommendations](https://developer.lsst.io/python/formatting.html#id1).

kfindeisen: I think calling it a "formatting recommendation" is a bit much, since it's not in the style guide and we're not required to use `black`. But I'll concede that the new commit message makes the motivation plenty clear. 👍

leeskelvin: :+1: Commit removed.

PR #142, opened 2023-06-15
DM-39681: Add LATISS pipelines to ap_pipe to be used by prompt processing

hsinfang:


pipelines/LATISS/ApPipe.yaml
```diff
@@ -0,0 +1,15 @@
...
+  - location: $AP_PIPE_DIR/pipelines/_ingredients/ApPipe.yaml
+    include:
+      - isr
+      - characterizeImage
+      - calibrate
+      - retrieveTemplate
+      - subtractImages
+      - detectAndMeasure
+      - transformDiaSrcCat
+      - diaPipe
```
kfindeisen, line 13: I'd recommend not enumerating these tasks -- they are subject to change, and it will surprise developers if only this pipeline breaks but the others (which only `exclude`) do not.

pipelines/LATISS/ApPipe.yaml
```diff
@@ -0,0 +1,15 @@
...
+parameters:
+  coaddName: deep
```
kfindeisen, line 15: This seems an odd thing to put in the main pipeline, since the correct value of `coaddName` is a function of the data the pipeline is run on. Is it expected that all LATISS coadds will be `deep`? If not, I recommend this be moved to lsst-dm/prompt_prototype#73.

kfindeisen (approved):
I'm a bit surprised that you could get the pipeline to run with only default configs, but I see that DRP doesn't have any LATISS-specific configs, either. Just a few comments about keeping the pipeline in this package as generic and forward-compatible as possible.

PR #143, opened 2023-06-25
DM-39467 Increase fake source density

BrunoSanchez:
This modifies the fake injection to increase the magnitude range and density of fakes. 
Setting new defaults for using fake-based metrics that require higher statistics. 

Probably density ends up being higher than needed. This would be a point of discussion before merging.

### DIFF
--- a/pipelines/HyperSuprimeCam/ApPipeWithFakes.yaml
+++ b/pipelines/HyperSuprimeCam/ApPipeWithFakes.yaml
@@ -19,8 +19,20 @@ imports:
       - detectAndMeasure
       - diaPipe
       - transformDiaSrcCat
+parameters:
+  coaddName: goodSeeing
+  fakesType: 'fakes_'
 
 tasks:
+  createFakes:
+    class: lsst.ap.pipe.createApFakes.CreateRandomApFakesTask
+    config:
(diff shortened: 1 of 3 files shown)


isullivan (approved):
Looks good, but please rebase to squash the four commits into a single commit.

BrunoSanchez:
Can't fix the check. I rebased but this still won't allow the merge in green.

timj:
Currently this branch has not been rebased. That's why it's still complaining. 

The error says:

> See https://developer.lsst.io/work/flow.html#pushing-code for detailed instructions.

pipelines/HyperSuprimeCam/ApPipeWithFakes.yaml
```diff
@@ -19,8 +19,20 @@ imports:
       - detectAndMeasure
       - diaPipe
       - transformDiaSrcCat
+parameters:
+  coaddName: goodSeeing
+  fakesType: 'fakes_'
```
kfindeisen, line 24: Sorry I missed this before, but why this override relative to `_ingredients/ApPipeWithFakes`? It will cause unexpected behavior if we ever need to change defaults.

PR #145, opened 2023-07-07
DM-36477: Remove ap_verify_hits2015 dataset

kfindeisen:
This PR fixes a large number of build problems and outdated information in the `ap_pipe` documentation, then updates the tutorial to use `ap_verify_ci_hits2015` instead of the obsolete `ap_verify_hits2015`.


rai-harshit (approved):
I assume there will be rebasing and squashing before the merging.

PR #147, opened 2023-07-26
DM-39858: Integrate new CalibrateImageTask with AP pipeline

parejkoj:


config/LSSTCam-imSim/calibrateImage.py
kfindeisen, line 1: Does this really belong in `ap_verify`, and not in `obs_lsst`? Since this is a new task, there's no reason we can't make the config RFC-775-compliant from the beginning.

pipelines/LsstCamImSim/ApPipeCalibrateImage.yaml
```diff
@@ -0,0 +1,26 @@
...
+#     (diaPipe configs should match the make_apdb.py configs)
+
+instrument: lsst.obs.lsst.LsstCamImSim
+imports:
+  - location: $AP_PIPE_DIR/pipelines/LsstCamImSim/ProcessCcdCalibrateImage.yaml
+  - location: $AP_PIPE_DIR/pipelines/_ingredients/ApPipeCalibrateImage.yaml
+    exclude:  # These tasks come from LsstCamImSim/ProcessCcd.yaml instead
+      - processCcd
+subsets:
+# apPipe must be redefined from $AP_PIPE_DIR/pipelines/_ingredients/ApPipe.yaml
```
kfindeisen, line 14:
```suggestion
# apPipe must be redefined from $AP_PIPE_DIR/pipelines/_ingredients/ApPipeCalibrateImage.yaml
```

pipelines/LsstCamImSim/ProcessCcdCalibrateImage.yaml
```diff
@@ -0,0 +1,27 @@
...
+      file: $AP_PIPE_DIR/config/LSSTCam-imSim/calibrateImage.py
+      # Do not integrate file and pipeline configs until DM-31047 fixed, to
+      # make it easier to check for changes on the obs side.
+      # Config file wipes out all pre-existing configs, so copy base pipeline
+      # config on top.
+      connections.astrometry_ref_cat: 'cal_ref_cat_2_2'
+      connections.photometry_ref_cat: 'cal_ref_cat_2_2'
+      python: >
+        config.astrometry_ref_loader.filterMap = {band: 'lsst_%s_smeared' % (band) for band in 'ugrizy'};
+        config.photometry_ref_loader.filterMap = {band: 'lsst_%s_smeared' % (band) for band in 'ugrizy'};
```
kfindeisen, line 27: This is a new task, and not susceptible to DM-31047 (especially since `calibrateImage.py` doesn't have the "reset to task default" hack).

pipelines/_ingredients/ApPipeCalibrateImage.yaml
```diff
@@ -0,0 +1,90 @@
...
+  - location: $AP_PIPE_DIR/pipelines/_ingredients/ProcessCcdCalibrateImage.yaml
+parameters:
+  # Pipeline configurable to run on both goodSeeing and deep templates, depending on dataset.
+  coaddName: goodSeeing
+tasks:
+  retrieveTemplate:  # For multi-tract difference imaging
+    class: lsst.ip.diffim.getTemplate.GetTemplateTask
+    config:
+      connections.bbox: initial_pvi.bbox
+      connections.wcs: initial_pvi.wcs
```
kfindeisen, line 28: I suppose it's outside the scope of this issue, but why does `calibrateImage` have outputs that are incompatible with the old `calibrate`? That will force _all_ pipelines (both AP and DRP) to have lengthy overrides of this sort.

pipelines/_ingredients/ApPipeCalibrateImage.yaml
```diff
@@ -0,0 +1,90 @@
...
+# e.g., $AP_PIPE_DIR/config/$CAMERA/someTask.py, or in a camera-specific,
+# pipeline, e.g., $AP_PIPE_DIR/pipelines/$CAMERA/ApPipe.yaml.
+
+imports:
+  - location: $AP_PIPE_DIR/pipelines/_ingredients/ProcessCcdCalibrateImage.yaml
+parameters:
+  # Pipeline configurable to run on both goodSeeing and deep templates, depending on dataset.
+  coaddName: goodSeeing
+tasks:
+  retrieveTemplate:  # For multi-tract difference imaging
```
kfindeisen, line 24: I think this comment is vestigial (it dates back to before there was a top-level `GetTemplateTask`).

pipelines/_ingredients/ApTemplateCalibrateImage.yaml
```diff
@@ -0,0 +1,80 @@
+description: The AP template building pipeline, with the new CalibrateImageTask.
+# Look in subdirectories of $AP_PIPE_DIR/pipelines to find customized pipelines
+# for each camera. Those pipelines import this general template-building pipeline.
+
+# WARNING: some camera-specific pipelines importing this pipeline presently
+# blow away all the configs that are set in this file.
+# To update a pipeline config prior to DM-35504, you MUST put it in either,
+# e.g., $AP_PIPE_DIR/config/$CAMERA/someTask.py, or in a camera-specific,
+# pipeline, e.g., $AP_PIPE_DIR/pipelines/$CAMERA/ApTemplate.yaml.
```
kfindeisen, line 9:
```suggestion
# pipeline, e.g., $AP_PIPE_DIR/pipelines/$CAMERA/ApTemplateCalibrateImage.yaml.
```

pipelines/_ingredients/ApTemplateCalibrateImage.yaml
```diff
@@ -0,0 +1,80 @@
...
+
+imports:
+  - location: $AP_PIPE_DIR/pipelines/_ingredients/ProcessCcd.yaml
```
kfindeisen, line 12:
```suggestion
  - location: $AP_PIPE_DIR/pipelines/_ingredients/ProcessCcdCalibrateImage.yaml
```

pipelines/_ingredients/ProcessCcdCalibrateImage.yaml
```diff
@@ -0,0 +1,20 @@
+description: ProcessCcd - A set of tasks to run when processing raw images.
+
+# This uses the new CalibrateImageTask, replacing characterizeImage and
+# calibrate.
+# WARNING: some camera-specific pipelines importing this pipeline presently
+# blow away all the configs that are set in this file.
+# To update a pipeline config prior to DM-35504, you MUST put it in either,
+# e.g., $AP_PIPE_DIR/config/$CAMERA/someTask.py, or in a camera-specific,
+# pipeline, e.g., $AP_PIPE_DIR/pipelines/$CAMERA/ProcessCcd.yaml.
```
kfindeisen, line 9:
```suggestion
# pipeline, e.g., $AP_PIPE_DIR/pipelines/$CAMERA/ProcessCcdCalibrateImage.yaml.
```

kfindeisen (changes requested):
The DM-31047 hacks appear to have been cargo-culted as-is, despite the fact that, as a task introduced after RFC-775, `calibrateImage` shouldn't have any DRP bias in its `obs_` configs. Then again, I _can't_ find any configs for it in `obs_lsst`... what's going on?

> Would it be better to put all the possibilities in the same ingredients files, and then have separate `XCalibrateImage.yaml` for the instrument-specific ones? Suggestions, please!

Sorry, but even after looking at the changes, I still don't understand this question. What are the "possibilities" you mention?

kfindeisen:
Also, I just noticed that the commit message mentions `ap_verify.py`, which is not an `ap_pipe` thing. 😉

pipelines/_ingredients/ApPipeCalibrateImage.yaml
kfindeisen: Looking at the `ap_verify` changes, I think it might be less disruptive to rename the `calibrateImage` outputs to the `calibrate` convention. As it is, a _lot_ of metrics need to be reconfigured.

config/LSSTCam-imSim/calibrateImage.py
parejkoj: Great, that's just what I was hoping to hear! I wasn't sure if we wanted to fix everything at once, or if we could migrate things one at a time. I'll happily move it over to obs_lsst (because it really is the generic settings you always want for imsim).

pipelines/_ingredients/ApPipeCalibrateImage.yaml
parejkoj: That was a deliberate choice following the design laid out in [DM-37584](https://jira.lsstcorp.org/browse/DM-37584). These are the names we want for these products eventually anyway, drp will mostly be making new tasks themselves per DMTN-172, and the outputs are different, though currently subtly: `initial_pvi` is photometrically calibrated while `calexp` is not, and `initial_stars_footprints_detector` is a much reduced list of sources compared with `src` (and will eventually only contain the footprints, and instead we'll have a parquet table for the sources themselves).

pipelines/_ingredients/ApTemplateCalibrateImage.yaml
parejkoj: Just to check: this warning re: DM-35504 still applies here? If we're adding a CalibrateImage override file to `obs_lsst/config/imsim` (which we want, because they're very specific to imsim), does follow the spirit comment given here, go against it, or is it orthogonal to this?

pipelines/_ingredients/ApTemplateCalibrateImage.yaml
parejkoj: Thinking about it more: I'm going to remove this new ApTemplate pipeline. We need to confirm that we can run something like DRP with the new task first, instead of just trying to make templates with it (which I haven't done on this ticket).

parejkoj:
>> Would it be better to put all the possibilities in the same ingredients files, and then have separate XCalibrateImage.yaml for the instrument-specific ones? Suggestions, please!

>Sorry, but even after looking at the changes, I still don't understand this question. What are the "possibilities" you mention?

For example, I could include this line in the ingredients file:

```
  calibrateImage: lsst.pipe.tasks.calibrateImage.CalibrateImageTask
```

 and then in the old files exclude it while the new files exclude `characterizeImage`/`calibrate`. I don't know that that actually gains us anything, since the ProcessCcd file isn't very complicated to begin with, but it was another possible approach I considered.

parejkoj:
> Also, I just noticed that the commit message mentions `ap_verify.py`, which is not an `ap_pipe` thing. 😉

I wrote it that way because ap_verify is the way we're going to start testing the new task. I can reword it to take out that reference?

pipelines/_ingredients/ApTemplateCalibrateImage.yaml
kfindeisen: Yes, it still applies. The comment is about the pipeline as a whole, and we still have to hack around some overrides for e.g. `makeWarp`.

kfindeisen:
> For example, I could include this line in the ingredients file:
> 
>     calibrateImage: lsst.pipe.tasks.calibrateImage.CalibrateImageTask
> 
> and then in the old files exclude it while the new files exclude characterizeImage/calibrate. I don't know that that actually gains us anything, since the ProcessCcd file isn't very complicated to begin with, but it was another possible approach I considered.

Got it. No, I'd rather not; part of my argument on [DM-40210](https://jira.lsstcorp.org/browse/DM-40210) was that the `_ingredients` pipeline is the first reference for "what does this pipeline do, anyway?", and having mutually exclusive tasks there would just make things confusing.

kfindeisen (approved):
Looks much better; main outstanding request is for forward-compatibility with future `ProcessCcd.yaml`.

pipelines/HyperSuprimeCam/ApTemplate.yaml
```diff
@@ -40,6 +40,8 @@ tasks:
       assembleStaticSkyModel.doSelectVisits: True
       connections.outputCoaddName: parameters.coaddName
 
+# TODO DM-40389: Remove this entire block once we are using calibrateImage in
+# the base _ingredients/ApTemplate.yaml.
```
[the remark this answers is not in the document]
parejkoj: Because those single frame tasks in `HyperSuprimeCam/ProcessCcd.yaml` will be gone: they're overriding tasks that will no longer exist.

pipelines/HyperSuprimeCam/ProcessCcd.yaml
```diff
@@ -1,3 +1,4 @@
+# TODO DM-40388: Remove this file once the new pipeline is vetted.
```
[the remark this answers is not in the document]
parejkoj: Ok, I've moved the comment down to the `task` block, as that whole thing can be removed.

pipelines/_ingredients/ApPipeCalibrateImage.yaml
```diff
@@ -0,0 +1,91 @@
...
+    class: lsst.ip.diffim.getTemplate.GetTemplateTask
+    config:
+      connections.bbox: initial_pvi.bbox
+      connections.wcs: initial_pvi.wcs
+      connections.coaddName: parameters.coaddName
+  subtractImages:
+    class: lsst.ip.diffim.subtractImages.AlardLuptonSubtractTask
+    config:
+      connections.science: initial_pvi
+      connections.sources: initial_stars_footprints_detector
```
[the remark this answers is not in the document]
parejkoj: I'm leaning towards the latter.

pipelines/_ingredients/ProcessCcdCalibrateImage.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+tasks:
+  isr: lsst.ip.isr.IsrTask
+  calibrateImage:
+    class: lsst.pipe.tasks.calibrateImage.CalibrateImageTask
+    config:
+      # TODO DM-39796: Temporary, until these become the repair defaults.
+      python: >
+        config.psf_repair.cosmicray.nCrPixelMax = 1000000;
+        config.psf_repair.cosmicray.cond3_fac2 = 0.4;
```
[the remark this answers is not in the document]
parejkoj: These are going to be pushed down into the repair overrides on DM-39796; they are defaults that we want for all instruments when doing CR repair.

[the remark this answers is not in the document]
kfindeisen: Yes, I understood that. I meant "do the temporary overrides belong in `ap_pipe`, and not in `CalibrateImageConfig`?"

pipelines/LsstCamImSim/ApPipeCalibrateImage.yaml
```diff
@@ -0,0 +1,10 @@
+# TODO DM-40388: This file will replace ApPipe.yaml, once the new pipeline is vetted.
+description: End to end AP pipeline specialized for ImSim.
+# (1) Execute `make_apdb.py`, e.g.,
+#     make_apdb.py -c db_url="sqlite:////project/user/association.db"
+# (2) Run this pipeline, setting appropriate diaPipe configs
+#     (diaPipe configs should match the make_apdb.py configs)
+
+instrument: lsst.obs.lsst.LsstCamImSim
+imports:
+  - location: $AP_PIPE_DIR/pipelines/_ingredients/ApPipeCalibrateImage.yaml
```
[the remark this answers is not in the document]
parejkoj: That seems a bit odd, since `_ingredients/ProcessCcdCalibrateImage.yaml` is already included by `_ingredients/ApPipeCalibrateImage.yaml`. Do we actually want that duplication in the instrument pipelines, or should we only add such if we know we need it (as in the current characterize/calibrate overrides)?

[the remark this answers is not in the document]
kfindeisen: Yes, it's redundant, but it would let you implement DM-40388 as a one-line substitution. The current `ApPipeCalibrateImage.yaml` is built around the fact that it _doesn't_ do any splicing, so adding it in later would be a major rewrite.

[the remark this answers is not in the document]
kfindeisen: Oh, I see what you meant -- "even if the new `ProcessCcd.yaml` is a no-change"? I think we do, just because the assumption that `<instrument>/ProcessCcd.yaml` doesn't have any tweaks is a brittle one.

pipelines/_ingredients/ProcessCcdCalibrateImage.yaml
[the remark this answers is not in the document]
parejkoj: I think so: this is the only pipeline that uses CalibrateImage right now, and I'm getting DM-39796 ready for review, so these lines won't last very long (and it saves me from making a change on pipe_tasks on this ticket).

pipelines/LsstCamImSim/ApPipeCalibrateImage.yaml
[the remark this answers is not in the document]
parejkoj: Wouldn't most such tweaks also require us to add subtask redefinitions to the instrument files anyway?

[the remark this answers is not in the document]
kfindeisen: I'm not sure I follow the argument; I was thinking of things like different configs. But maybe in the post-RFC-775 world, configs that need to be both AP- and instrument-specific are rarer?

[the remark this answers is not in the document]
parejkoj: I was thinking of things like the DECam overrides that exclude isr and thus require redefining the `processCcd` subset in `ApPipe.yaml`.

But yes, once DM-39796 is merged, I'd expect instrument-specific AP overrides to be very rare, and be almost exclusively in ISR (as in DECam).

PR #148, opened 2023-08-03
DM-40210: Clean up ap_pipe and ap_verify pipelines

kfindeisen:
This PR removes a large amount of duplicate pipeline configuration, particularly base (`_ingredients`) configs and tasks copied to instrument-specific pipelines. This leaves the pipelines cleaner and with clearer responsibilities, and will prevent changes to the base pipeline from unexpectedly not working.

Unfortunately, not all of the duplicate configs in the instrument pipelines could be removed, as some are still needed as workarounds for DRP-specific settings/plugins lurking in the `obs` package configs.

### DIFF
--- a/pipelines/DarkEnergyCamera/ApPipeWithFakes.yaml
+++ b/pipelines/DarkEnergyCamera/ApPipeWithFakes.yaml
@@ -3,8 +3,7 @@ description: DECam AP Pipeline with synthetic/fake sources. Templates are inputs
 # (0) Ensure median calibration products and template coadds exist for the data being processed
 # (1) Run $AP_PIPE_DIR/pipelines/DarkEnergyCamera/RunIsrForCrosstalkSources.yaml
 # (2) Execute `make_apdb.py`, e.g.,
-#     make_apdb.py -c isolation_level=READ_UNCOMMITTED
-#                  -c db_url="sqlite:////project/user/association.db"
+#     make_apdb.py -c db_url="sqlite:////project/user/association.db"
 # (3) Run this pipeline, setting appropriate diaPipe configs
 #     (diaPipe configs must match the make_apdb.py configs)
 
--- a/pipelines/DarkEnergyCamera/ProcessCcd.yaml
+++ b/pipelines/DarkEnergyCamera/ProcessCcd.yaml
@@ -24,8 +24,6 @@ tasks:
       file: $AP_PIPE_DIR/config/DECam/calibrate.py
       # Do not integrate file and pipeline configs until DM-31047 fixed, to
       # make it easier to check for changes on the obs side.
-      photoCal.match.referenceSelection.magLimit.fluxField: 'i_flux'
-      photoCal.match.referenceSelection.magLimit.maximum: 22.0
 subsets:
 # processCcd must be redefined from
 # $AP_PIPE_DIR/pipelines/_ingredients/ProcessCcd.yaml because its isr was
--- a/pipelines/DarkEnergyCamera/ProcessCcd.yaml
+++ b/pipelines/DarkEnergyCamera/ProcessCcd.yaml
@@ -11,15 +11,19 @@ tasks:
   characterizeImage:
     class: lsst.pipe.tasks.characterizeImage.CharacterizeImageTask
     config:
+      # HACK: workaround for DM-31047, remove once we're no longer loading
+      # DRP-specific obs configs.
       file: $AP_PIPE_DIR/config/DECam/characterizeImage.py
       # Config file wipes out all pre-existing configs, so copy base pipeline
       # config on top.
   calibrate:
     class: lsst.pipe.tasks.calibrate.CalibrateTask
     config:
+      # HACK: workaround for DM-31047, remove once we're no longer loading
+      # DRP-specific obs configs.
       file: $AP_PIPE_DIR/config/DECam/calibrate.py
-      # Do not integrate file and pipeline configs until obs config files are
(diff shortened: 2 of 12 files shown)


pipelines/DarkEnergyCamera/ProcessCcd.yaml
```diff
@@ -24,8 +24,6 @@ tasks:
       file: $AP_PIPE_DIR/config/DECam/calibrate.py
       # Do not integrate file and pipeline configs until DM-31047 fixed, to
       # make it easier to check for changes on the obs side.
-      photoCal.match.referenceSelection.magLimit.fluxField: 'i_flux'
-      photoCal.match.referenceSelection.magLimit.maximum: 22.0
```
mrawls, line 24: I agree these can safely be removed because they are set via `config/DECam/calibrate.py` instead

pipelines/_ingredients/ApPipeWithFakes.yaml
```diff
@@ -41,8 +41,6 @@ tasks:
       connections.fakesType: parameters.fakesType
       insertFakes.doSubSelectSources: True
       insertFakes.select_col: 'isVisitSource'
-      calibrate.photoCal.match.referenceSelection.magLimit.fluxField: i_flux
-      calibrate.photoCal.match.referenceSelection.magLimit.maximum: 22.0
```
mrawls, line 45: I appreciate that you don't want these in the generic pipeline. Since you are removing them here, they ought to land in each camera's `processVisitFakes` pipeline instead. In practice, both HSC and DECam use refcats that have the `i_flux` flux field, and this is the best proxy config available for "only use sources with a decent signal-to-noise-ratio for photometric calibration."

pipelines/HyperSuprimeCam/ApPipeWithFakes.yaml
```diff
@@ -34,5 +34,3 @@ tasks:
       connections.fakesType: parameters.fakesType
       insertFakes.doSubSelectSources: True
       insertFakes.select_col: 'isVisitSource'
-      calibrate.photoCal.match.referenceSelection.magLimit.fluxField: i_flux
-      calibrate.photoCal.match.referenceSelection.magLimit.maximum: 22.0
```
mrawls, line 48: Please don't remove these, unless you add them to `config/HSC/processCcdWithFakes.py` instead

pipelines/DarkEnergyCamera/ApPipeWithFakes.yaml
```diff
@@ -3,8 +3,7 @@ description: DECam AP Pipeline with synthetic/fake sources. Templates are inputs
 # (0) Ensure median calibration products and template coadds exist for the data being processed
 # (1) Run $AP_PIPE_DIR/pipelines/DarkEnergyCamera/RunIsrForCrosstalkSources.yaml
 # (2) Execute `make_apdb.py`, e.g.,
-#     make_apdb.py -c isolation_level=READ_UNCOMMITTED
-#                  -c db_url="sqlite:////project/user/association.db"
+#     make_apdb.py -c db_url="sqlite:////project/user/association.db"
 # (3) Run this pipeline, setting appropriate diaPipe configs
 #     (diaPipe configs must match the make_apdb.py configs)
```
mrawls, line 8: Missing: your favorite two `processVisitFakes.calibrate.photoCal.match[...]` configs, either in this file or in `config/DECam/processCcdWithFakes.py`

mrawls (commented):
Thanks for these cleanups.

For human readability, I did like the reminder to config the APDB URL in the camera-specific pipelines, but I can live without it, since it's not really where documentation belongs.

I believe the two calibrate photoCal config lines should still appear each time a camera-specific pipeline uses calibrateTask. This is extra confusing with our current fakes system, because processCcdVisitFakes calls calibrateTask a second time and can't re-use the original configs.

pipelines/_ingredients/ApPipeWithFakes.yaml
kfindeisen: Except that a lot of the confusion surrounding these overrides came from the fact that `processVisitFakes` is _not_ supposed to run `photoCal`, and ignores these fields -- which is why none of the `processCcdWithFakes.py` files bothered setting them. So I really think the best solution is to not mention them in the pipelines at all, and trust that the config files are already doing the (generically) right thing.

kfindeisen:
> For human readability, I did like the reminder to config the APDB URL in the camera-specific pipelines, but I can live without it, since it's not really where documentation belongs.

Sorry, I probably should have asked about this. Were those more visible than the detailed instructions at the top of the file?

pipelines/_ingredients/ApPipeWithFakes.yaml
mrawls: Now that I appreciate this nuance, I retract my request to leave these configs anywhere in "calibrateTask Take 2" that happens inside processCcdWithFakes.

pipelines/HyperSuprimeCam/ApPipeWithFakes.yaml
mrawls: Never mind, see above!

mrawls:
The APDB URL config is the kind of thing that I find hard to format correctly without boilerplate, and I've been burned many times by forgetting to configure it when running a pipeline. That said, it fails fast with a friendly and helpful error message when this happens, and there are many places besides the actual pipeline to provide the config. So I'm OK with just the instructions at the top of the file.

mrawls (approved):

PR #149, opened 2023-08-26
DM-39796: Update task config defaults to LSST values

parejkoj:


mrawls (approved):
This looks self-consistent as best I can tell. It would be nice if the commits were more descriptive (e.g., "this is set in pipe_tasks instead") and/or squashed into one big "Remove now-defaulted overrides" commit with the relevant descriptions bundled together there.

config/DECam/calibrate.py
```diff
@@ -47,9 +47,6 @@
         matchConfig.matcher.maxMatchDistArcSec = 2.0
         matchConfig.sourceSelector.active.excludePixelFlags = False
 
-# Demand astrometry and photoCal succeed
-config.requireAstrometry = True
-config.requirePhotoCal = True
```
kfindeisen, line 59: I do not understand the logic behind this commit. **We**, specifically, need astrometry and photometry, regardless of what the task default is or how any upstream pipelines are configured.

PR #151, opened 2023-10-31
DM-41498: Reorganize the pipeline config Prompt Processing uses 

hsinfang:


pipelines/LATISS/ApPipe.yaml
```diff
@@ -2,3 +2,16 @@ description: AP pipeline specialized for LATISS
 instrument: lsst.obs.lsst.Latiss
```
kfindeisen, line 2: I recommend updating the description to mention that this pipeline is optimized for Prompt Processing, especially if the comments use PP's limitations to justify the config overrides.

kfindeisen (approved):
Looks good, but please make sure it's clear to the reader that the pipeline is designed around Prompt Processing (unlike the other pipeline variants, which are designed around "batch" testing).

PR #154, opened 2023-11-24
DM-41838: Update maxScaledSizeScatter thresholds for LSSTCam-imSim

laurenam:


kfindeisen (commented):
Thanks for keeping this in sync with `obs_lsst`! Hopefully [DM-31047](https://jira.lsstcorp.org/browse/DM-31047) will get done one of these days...

As for testing, this file is only used by `LsstCamImSim/ApTemplate.yaml`, so the only way to test it would be to manually make some coadds.

laurenam:
Just to make sure.  Do you approve merging this branch?  (The PR has not been officially approved...)

kfindeisen:
I don't see any AP-specific problems, but I'm not qualified to comment on the numbers. So "cautious yes"?

PR #157, opened 2024-01-29
DM-40798: Exclude sky sources from RB classification

rai-harshit:


pipelines/_ingredients/ApPipe.yaml
```diff
@@ -73,8 +79,6 @@ contracts:
       diaPipe.connections.ConnectionsClass(config=diaPipe).exposure.name
   - detectAndMeasure.connections.ConnectionsClass(config=detectAndMeasure).subtractedMeasuredExposure.name ==
       transformDiaSrcCat.connections.ConnectionsClass(config=transformDiaSrcCat).diffIm.name
-  - detectAndMeasure.connections.ConnectionsClass(config=detectAndMeasure).diaSources.name ==
-      transformDiaSrcCat.connections.ConnectionsClass(config=transformDiaSrcCat).diaSourceCat.name
```
kfindeisen, line 77: Could you add contracts that link `filterDiaSrcCat` to `detectAndMeasure`(?) and `transformDiaSrcCat`?

kfindeisen: Though given the discussion on RFC-987, perhaps you'd deliberately want the input to `transformDiaSrcCat` to be ambiguous...

kfindeisen: (Please disregard my previous post -- a modified version of `ApPipe` that excluded `filterDiaSrcCat` would also lose the contracts that apply to it, so there's no need to specifically support the possibility.)

rai-harshit: @kfindeisen I've modified the contract as you suggested and have also rearranged the tasks in their execution order. Please let me know if things look good now.

kfindeisen: Yes, looks good. I'll keep my nose out of somebody else's review now. 🙂

isullivan (approved):

PR #163, opened 2024-02-15
DM-42419: Break up ApVerify into pure with- and without-fakes pipelines

kfindeisen:
This PR removes the disruptive "WithFakes" suffixes from tasks in `ApPipeWithFakes`. It must be merged together with lsst/ap_verify#213.


BrunoSanchez (approved):
Looks good. 👍

PR #164, opened 2024-02-20
DM-42220: Incorporate ModelPackage Butler datasets into Prompt Processing

kfindeisen:
This PR removes the pre-emptive blocker that prevented `rbClassify` from being run in LATISS (specifically, Prompt Processing). It must be merged after lsst-dm/prompt_processing#125.

### DIFF
--- a/pipelines/LATISS/ApPipe.yaml
+++ b/pipelines/LATISS/ApPipe.yaml
@@ -5,13 +5,3 @@ description: |
 instrument: lsst.obs.lsst.Latiss
 imports:
   - location: $AP_PIPE_DIR/pipelines/_ingredients/ApPipe.yaml
-    exclude:
-      # TODO: prompt_prototype does not yet support R/B analysis: DM-42220.
-      - rbClassify
-
-tasks:
-  transformDiaSrcCat:
-    class: lsst.ap.association.TransformDiaSourceCatalogTask
-    config:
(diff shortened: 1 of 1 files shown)


pipelines/LATISS/ApPipe.yaml
```diff
@@ -5,13 +5,3 @@ description: |
...
-    exclude:
-      # TODO: prompt_prototype does not yet support R/B analysis: DM-42220.
-      - rbClassify
-
-tasks:
-  transformDiaSrcCat:
-    class: lsst.ap.association.TransformDiaSourceCatalogTask
-    config:
-      # TODO: prompt_prototype does not yet support R/B analysis: DM-42220.
-      doIncludeReliability: False
```
hsinfang, line 17: Should `pipelines/LSSTComCamSim/ApPipe.yaml` be updated too, assuming the model can be added to `/repo/ops-rehearsal-3-prep`?

hsinfang (approved):

pipelines/LATISS/ApPipe.yaml
kfindeisen: In theory, yes. But everyone seemed skeptical that that was a good idea. I can add it now, or save it for a later issue after we've seen how it works on LATISS-prod...

hsinfang: Leaving it for later is totally fine

PR #168, opened 2024-03-07
DM-40400: Create total AP pipeline timing metric

kfindeisen:
This PR implements `PipelineTimingMetricTask`, which allows timing of pipelines that are linear or close to linear. It's designed to work on `ApPipe` or closely related pipelines (e.g., `ApVerify`), but is not a general-purpose tool.

### DIFF
--- a/python/lsst/ap/pipe/metrics.py
+++ b/python/lsst/ap/pipe/metrics.py
@@ -25,18 +25,20 @@
 
 __all__ = [
     "ApFakesCompletenessMetricTask", "ApFakesCompletenessMetricConfig",
-    "ApFakesCountMetricTask", "ApFakesCountMetricConfig"
+    "ApFakesCountMetricTask", "ApFakesCountMetricConfig",
+    "PipelineTimingMetricTask", "PipelineTimingMetricConfig",
 ]
 
 import astropy.units as u
+from datetime import datetime
 import numpy as np
 
 import lsst.pex.config as pexConfig
-from lsst.pipe.base import Struct
+from lsst.pipe.base import Struct, NoWorkFound
 import lsst.pipe.base.connectionTypes as connTypes
 from lsst.pipe.tasks.insertFakes import InsertFakesConfig
-from lsst.verify import Measurement
-from lsst.verify.tasks import MetricTask, MetricComputationError
+from lsst.verify import Measurement, Datum
+from lsst.verify.tasks import AbstractMetadataMetricTask, MetricTask, MetricComputationError
 
 
 class ApFakesCompletenessMetricConnections(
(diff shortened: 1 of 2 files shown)


tests/test_metrics.py
```diff
@@ -190,6 +195,114 @@ def testValidEmpty(self):
...
+        try:
+            result = task.run(self.startTask.getFullMetadata(), self.endTask.getFullMetadata())
+        except lsst.pipe.base.NoWorkFound:
+            # Correct behavior
+            pass
+        else:
+            # Alternative correct behavior
+            testUtils.assertValidOutput(task, result)
+            meas = result.measurement
+            self.assertIsNone(meas)
```
parejkoj, line 277: This really confuses me: how can the `else` here be valid, since I read `targetStart = "first.doProcess"` as forcing a `NoWorkFound`. Similarly below.

Maybe some docstrings on these two tests would help clarify the intended behavior?

tests/test_metrics.py
```diff
@@ -190,6 +195,114 @@ def testValidEmpty(self):
...
+                   if "EndUtc" in key]
+        for key in endKeys:
+            metadata[key] = 42
+
+        with self.assertRaises(lsst.verify.tasks.MetricComputationError):
+            self.task.run(self.startTask.getFullMetadata(), metadata)
+
+
 # Hack around unittest's hacky test setup system
 del MetricTaskTestCase
```
parejkoj, line 307: I know it's outside ticket, but this wouldn't be necessary if `MetricTaskTestCase` wasn't derived from `lsst.utils.tests.TestCase`. The concrete classes that implement the tests would have to explicitly derive from that, but it would prevent unittest from thinking the ABC is a test.

tests/test_metrics.py
```diff
@@ -190,6 +195,114 @@ def testValidEmpty(self):
...
+        except lsst.pipe.base.NoWorkFound:
+            # Correct behavior
+            pass
+        else:
+            # Alternative correct behavior
+            testUtils.assertValidOutput(task, result)
+            meas = result.measurement
+            self.assertIsNone(meas)
+
+    def testBadlyTypedKeys(self):
```
parejkoj, line 294: The Task itself catches both TypeError and ValueError; this only checks the former, and it's not clear to me why the latter is necessary. Unfortunately, `datetime.fromisoformat` doesn't list what it could raise, that I can find. Catching two exceptions in the test above made me wonder what the examples were to trigger each of them.

parejkoj (commented):
Handful of comments, mostly about documentation, plus a couple broader questions that don't necessarily need to be answered here.

python/lsst/ap/pipe/metrics.py
```diff
@@ -187,3 +189,133 @@ def run(self, matchedFakes, band):
...
+
+class PipelineTimingMetricConfig(MetricTask.ConfigClass, pipelineConnections=PipelineTimingMetricConnections):
+    # Don't include the dimensions hack that MetadataMetricConfig has; unlike
+    # TimingMetricTask, this task is not designed to be run on multiple
+    # pipelines.
+    targetStart = pexConfig.Field(
+        dtype=str,
+        doc="The method to profile on the starting task, optionally prefixed by "
+            "one or more tasks in the format of "
+            "`lsst.pipe.base.Task.getFullMetadata()`.")
```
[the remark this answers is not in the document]
kfindeisen: Maybe it's because I'm not a native English speaker, but I can't parse the proposed replacement.

"The method to take as the starting point on the starting task"? Ignoring the redundant redundancy...

python/lsst/ap/pipe/metrics.py
```diff
@@ -187,3 +189,133 @@ def run(self, matchedFakes, band):
...
+class PipelineTimingMetricTask(AbstractMetadataMetricTask):
+    """A Task that computes a wall-clock time for an entire pipeline, using
+    metadata produced by the `lsst.utils.timer.timeMethod` decorator.
+
+    Parameters
+    ----------
+    args
+    kwargs
+        Constructor parameters are the same as for
+        `lsst.verify.tasks.MetricTask`.
```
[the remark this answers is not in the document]
kfindeisen: I would assume that if no parameters are listed, then the constructor takes no parameters. In any case, there's no "natural" link to `MetricTask` (its great-grandparent), even though it would be useful for anyone looking at an implementation class.

tests/test_metrics.py
kfindeisen: [The spec on `MetricTask` says that both of these are allowed](https://pipelines.lsst.io/v/daily/modules/lsst.verify/tasks/lsst.verify.tasks.MetricTask.html#error-handling). I do not write tests that restate the implementation.

tests/test_metrics.py
kfindeisen: Yes, but it would cause other problems (like code in one class being written against fields or methods in a completely different class). This is the lesser evil.

tests/test_metrics.py
kfindeisen: I'll admit I adapted the code from the existing `TimingMetricTask` and don't remember, but `ValueError` seems a reasonable exception for a string that is _not_ an ISO date.

python/lsst/ap/pipe/metrics.py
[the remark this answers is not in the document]
parejkoj: Oh, yes, that's actually much more clear.

python/lsst/ap/pipe/metrics.py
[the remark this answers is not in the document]
parejkoj: Ah, the class hierarchy wasn't fully clear to me.

tests/test_metrics.py
parejkoj: I guess I'm confused about the test behavior, since only one of those two cases is true in the class as-written. Or is the idea that it could be re-implemented with the other behavior and you don't want the test to have to change?

kfindeisen: The idea is that the test should not depend on the implementation _at all_; it should be expressed entirely in terms of correct (or at least acceptable) behavior. Either the class works as advertised, or it doesn't.

Customizing the "test" to match the internal code, and having it mirror any changes, would mean it's no longer testing anything; it would be like reviewing your own code and saying "Yes, I agree with me!"

python/lsst/ap/pipe/metrics.py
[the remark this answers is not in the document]
kfindeisen: Also replaced "one or more tasks" with "one or more task names", hopefully that reduces confusion.

PR #169, opened 2024-03-08
DM-41108: Add analysis_tools afterburner

isullivan:
Adds analysis_tools metrics to the ApPipe and ApPipeWithFakes pipelines.

### DIFF
--- a/pipelines/_ingredients/ApPipe.yaml
+++ b/pipelines/_ingredients/ApPipe.yaml
@@ -16,7 +16,8 @@ description: End to end Alert Production pipeline.
 # pipeline, e.g., $AP_PIPE_DIR/pipelines/$CAMERA/ApPipeCalibrateImage.yaml.
 
 imports:
-  - location: $AP_PIPE_DIR/pipelines/_ingredients/ProcessCcd.yaml
+  - $AP_PIPE_DIR/pipelines/_ingredients/ProcessCcd.yaml
+  - $ANALYSIS_TOOLS_DIR/pipelines/apDetectorVisitQualityCore.yaml
 parameters:
   # Pipeline configurable to run on both goodSeeing and deep templates, depending on dataset.
   coaddName: goodSeeing
@@ -74,6 +75,9 @@ subsets:
       - rbClassify
       - transformDiaSrcCat
       - diaPipe
+      - analyzeAssocDiaSrcCore
+      - analyzeTrailedDiaSrcCore
+      - diffimTaskCore
     description: >
       An alias of ApPipe to use in higher-level pipelines.
 contracts:
--- a/pipelines/_ingredients/ApPipeWithFakes.yaml
+++ b/pipelines/_ingredients/ApPipeWithFakes.yaml
@@ -16,6 +16,7 @@ imports:
   # The existing Fakes insertion system is not currently compatible with the
   # new CalibrateImageTask, so the fakes pipeline still is on the old tasks.
   - location: $AP_PIPE_DIR/pipelines/_ingredients/ProcessCcdCalibrate.yaml
+  - location: $ANALYSIS_TOOLS_DIR/pipelines/apDetectorVisitQualityCore.yaml
 parameters:
   coaddName: goodSeeing
   fakesType: 'fakes_'
@@ -110,6 +111,9 @@ subsets:
       - transformDiaSrcCat
       - diaPipe
       - fakesMatch
+      - analyzeAssocDiaSrcCore
+      - analyzeTrailedDiaSrcCore
+      - diffimTaskCore
     description: >
       The AP pipeline with fakes. Requires apPipe and prepareFakes subsets.
 
--- a/pipelines/DECam/ApPipe.yaml
+++ b/pipelines/DECam/ApPipe.yaml
@@ -24,5 +24,8 @@ subsets:
       - rbClassify
       - transformDiaSrcCat
       - diaPipe
+      - analyzeAssocDiaSrcCore
+      - analyzeTrailedDiaSrcCore
+      - diffimTaskCore
     description: >
       An alias of ApPipe to use in higher-level pipelines.
--- a/ups/ap_pipe.table
+++ b/ups/ap_pipe.table
@@ -11,6 +11,7 @@ setupRequired(pipe_base)
 setupRequired(pipe_tasks)
 
 setupRequired(ap_association)
+setupRequired(analysis_tools)
 
 setupRequired(meas_transiNet)
 


pipelines/_ingredients/ApPipe.yaml
```diff
@@ -16,7 +16,8 @@ description: End to end Alert Production pipeline.
 # pipeline, e.g., $AP_PIPE_DIR/pipelines/$CAMERA/ApPipeCalibrateImage.yaml.
 
 imports:
-  - location: $AP_PIPE_DIR/pipelines/_ingredients/ProcessCcd.yaml
+  - $AP_PIPE_DIR/pipelines/_ingredients/ProcessCcd.yaml
+  - $ANALYSIS_TOOLS_DIR/pipelines/apDetectorVisitQualityCore.yaml
```
kfindeisen, line 20: Please add `analysis_tools` to `ups/ap_pipe.table`. Otherwise, setting up `ap_pipe` won't let you use its pipelines.

pipelines/_ingredients/ApPipe.yaml
```diff
@@ -74,6 +75,10 @@ subsets:
       - rbClassify
       - transformDiaSrcCat
       - diaPipe
+  qaMetrics:
+    subset:
+      - analyzeAssocDiaSrcCore
+      - diffimTaskCore
     description: >
       An alias of ApPipe to use in higher-level pipelines.
```
kfindeisen, line 83: The subset definition got a bit garbled; the description belongs with `apPipe`.
```suggestion
    description: >
      An alias of ApPipe to use in higher-level pipelines.
  qaMetrics:
    subset:
      - analyzeAssocDiaSrcCore
      - diffimTaskCore
```

pipelines/_ingredients/ApPipe.yaml
```diff
@@ -74,6 +75,10 @@ subsets:
       - rbClassify
       - transformDiaSrcCat
       - diaPipe
+  qaMetrics:
+    subset:
+      - analyzeAssocDiaSrcCore
+      - diffimTaskCore
```
kfindeisen, line 81: Why not define `qaMetrics` in `apDetectorVisitQualityCore.yaml`? Does `analysis_tools` have rules/conventions against it?

Is this subset supposed to exclude `analyzeTrailedDiaSrcCore`?

kfindeisen: Please include all tasks in `apPipe` so that the alias behaves correctly in higher-level pipelines like `ApVerify`. Unfortunately, Middleware doesn't want subsets to be included as subset members, so you'll have to do it manually.

pipelines/_ingredients/ApPipeWithFakes.yaml
```diff
@@ -110,6 +111,10 @@ subsets:
       - transformDiaSrcCat
       - diaPipe
       - fakesMatch
+  qaMetrics:
+    subset:
+      - analyzeAssocDiaSrcCore
+      - diffimTaskCore
     description: >
       The AP pipeline with fakes. Requires apPipe and prepareFakes subsets.
```
kfindeisen, line 119:
```suggestion
    description: >
      The AP pipeline with fakes. Requires apPipe and prepareFakes subsets.
  qaMetrics:
    subset:
      - analyzeAssocDiaSrcCore
      - diffimTaskCore
```

pipelines/_ingredients/ApPipe.yaml
kfindeisen: If you could add a description for `qaMetrics`, of course, that would be great.

kfindeisen (changes requested):
The subset definitions seem to have gotten mixed up and incomplete; I suspect this may be the root cause of the problems you encountered on lsst/ap_verify#217.

pipelines/_ingredients/ApPipeWithFakes.yaml
kfindeisen: The fakes definition of `apPipe` seems to be broken, and to have been previously worked around in the specialized `ApVerifyWithFakes` by using `processCcd` in place of `apPipe`. Not sure if you'd want to fix that, or have me do it, but it would probably be easiest to address it _before_ adding the QA metrics.

isullivan: I removed the `qaMetrics` subset and put the metrics in `apPipe`. That seems to fix the import problems you pointed out elsewhere. I have verified that a local `ap_verify` run of `ApPipeWithFakes.yaml` on the `ap_verify_ci_hits2015` works.

kfindeisen: Ok. Just to be clear, I meant for the new tasks to be added to `apPipe` _in addition to_ `qaMetrics`, not instead of. A `qaMetrics` subset might be useful (we don't have one in `ap_verify` because the way metrics are split among different files makes defining it impractical).

pipelines/_ingredients/ApPipe.yaml
```diff
@@ -80,6 +80,14 @@ subsets:
...
     description: >
       An alias of ApPipe to use in higher-level pipelines.
+  promptQaMetrics:
+    subset:
+      - analyzeAssocDiaSrcCore
+      - analyzeTrailedDiaSrcCore
+      - diffimTaskCore
+    description: >
+      QA metrics to run in Prompt Processing.
+
```
kfindeisen, line 90: I don't think you need to redefine this; since you're not excluding any of these tasks on import, the definition from lsst/analysis_tools#218 should carry over.

pipelines/_ingredients/ApPipeWithFakes.yaml
kfindeisen: Filed [DM-43349](https://rubinobs.atlassian.net/browse/DM-43349) to deal with the preexisting bad `apPipe` definition.

kfindeisen (approved):
Looks clean, thanks!

pipelines/_ingredients/ApPipe.yaml
isullivan: This addition was in response to your comment "A qaMetrics subset might be useful...". Did I misinterpret your request? You are correct that everything works without this extra subset definition.

kfindeisen: Sorry, when I wrote that I didn't realize you were planning to move (had already moved?) the definition to `analysis_tools`, as previously suggested.

TLDR: it might be useful to have a definition, but there's definitely no need to define it in more than one place.

PR #172, opened 2024-03-23
DM-43476: Fix broken pipeline definitions with new test

parejkoj:


tests/test_pipelines.py
```diff
@@ -0,0 +1,68 @@
...
+        files = glob.glob(os.path.join(self.path, "_ingredients/*.yaml"))
+        for file in files:
+            if "ApTemplate" in file:
+                # Our ApTemplate definition cannot be tested here because it
+                # depends on drp_tasks, which we cannot make a dependency here.
+                continue
+            with self.subTest(file):
+                pipeline = lsst.pipe.base.Pipeline.from_uri(file)
+                if "apPipe" in pipeline.subsets:
+                    pipeline.addConfigPython("diaPipe", "config.apdb.db_url='sqlite:tests/data/test_apdb.db'")
```
kfindeisen, line 53: Why not:
```suggestion
                    pipeline.addConfigPython("diaPipe", "config.apdb.db_url='sqlite://'")
```
It's valid, and guarantees no side effects.

natelust (approved):
Looks fine to me

tests/test_pipelines.py
natelust: I’m confused on why this needs to be a python config block rather than just a regular config override

tests/test_pipelines.py
```diff
@@ -0,0 +1,68 @@
...
+    """Tests of the self-consistency of our pipeline definitions.
+    """
+    def setUp(self):
+        self.path = os.path.join(lsst.utils.getPackageDir("ap_pipe"), "pipelines")
+
+    def test_pipelines(self):
+        """Test that each pipeline definition file in `_ingredients/` can be
+        used to build a graph.
+        """
+        files = glob.glob(os.path.join(self.path, "_ingredients/*.yaml"))
```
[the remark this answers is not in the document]
parejkoj: I can't test the instrument-specific pipelines, because ap_pipe doesn't depend on any instrument packages.

tests/test_pipelines.py
parejkoj: I didn't know that was an option: I see it now as `addConfigOverride`, which is definitely clearer.

PR #175, opened 2024-04-05
DM-43711: Add contracts for analysis_tools plots

isullivan:
Note that most analysis_tools Tasks require the input connections to be named 'data'

### DIFF
--- a/pipelines/_ingredients/ApPipe.yaml
+++ b/pipelines/_ingredients/ApPipe.yaml
@@ -122,6 +122,9 @@ contracts:
   - contract: detectAndMeasure.connections.ConnectionsClass(config=detectAndMeasure).subtractedMeasuredExposure.name ==
               diaPipe.connections.ConnectionsClass(config=diaPipe).diffIm.name
     msg: "detectAndMeasure.subtractedMeasuredExposure != diaPipe.diffIm"
+  - contract: detectAndMeasure.connections.ConnectionsClass(config=detectAndMeasure).spatiallySampledMetrics.name ==
+              diffimTaskPlots.connections.ConnectionsClass(config=diffimTaskPlots).data.name
+    msg: "detectAndMeasure.spatiallySampledMetrics != diffimTaskPlots.data"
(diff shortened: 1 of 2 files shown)


pipelines/_ingredients/ApPipe.yaml
```diff
@@ -135,3 +135,6 @@ contracts:
   - contract: transformDiaSrcCat.connections.ConnectionsClass(config=transformDiaSrcCat).diaSourceTable.name ==
               diaPipe.connections.ConnectionsClass(config=diaPipe).diaSourceTable.name
     msg: "transformDiaSrcCat.diaSourceTable != diaPipe.diaSourceTable"
+  - contract: detectAndMeasure.connections.ConnectionsClass(config=detectAndMeasure).spatiallySampledMetrics.name ==
+              diffimTaskPlots.connections.ConnectionsClass(config=diffimTaskPlots).data.name
+    msg: "detectAndMeasure.spatiallySampledMetrics != diffimTaskPlots.spatiallySampledMetrics"
```
kfindeisen, line 140: So that people don't get lost when making future edits, please keep the contracts sorted as described in the comment. In this case, it would go with the other `detectAndMeasure` contracts, probably at the end (though `diffimTaskPlots` subverts the assumption of a linear pipeline...).

kfindeisen: Can you also add contracts for `analyzeAssocDiaSrcCore`, `analyzeTrailedDiaSrcCore`, and `diffimTaskCore`? The same bug could potentially happen there.

kfindeisen (approved):
Contract itself looks good, but please keep the file in order so that it's easy to find later.

pipelines/_ingredients/ApPipe.yaml
isullivan: Yes, though `diffimTaskCore` is weird because it reads the task metadata instead of the output data products. That does not get written with any of our parameters like `fakesType` or `coaddName`.

kfindeisen: Good point, a contract on the metadata would be redundant with the task labels already imposed by the pipeline.

PR #186, opened 2024-06-05
DM-44678: Set kernelSize explicitly

arunkannawadi:


TallJimbo (approved):
I think you could also add a contract to the `ApTemplate.yaml` similar to the one you've added to the DRP ingredients pipeline.

PR #187, opened 2024-06-06
DM-43856: Use sdm_schemas to set the datatypes of columns

isullivan:


ebellm (approved):

pipelines/_ingredients/ApPipe.yaml
```diff
@@ -58,6 +58,7 @@ tasks:
       doRemoveSkySources: True
       connections.coaddName: parameters.coaddName
       doIncludeReliability: True  # Output from rbClassify
+      doUseApdbSchema: True
```
kfindeisen, line 61: Please justify all new config overrides. 🙂 (Especially since in this case the flag name is not very informative.)

isullivan: I've added a comment, but I would also welcome suggestions for a better flag name!

PR #189, opened 2024-07-16
DM-42576: Write task to compute predicted positions for an upcoming visit

Gerenjie:
Adding mpSkyEphemerisQuery, which replaces SkyBot in querying mpSky for ssObjects in the pre-load step.

### DIFF
--- a/pipelines/DECam/ApPipe.yaml
+++ b/pipelines/DECam/ApPipe.yaml
@@ -25,6 +25,7 @@ subsets:
       - rbClassify
       - transformDiaSrcCat
       - getRegionTimeFromVisit
+      - mpSkyEphemerisQuery
       - diaPipe
       - analyzeAssocDiaSrcCore
       - analyzeTrailedDiaSrcCore
@@ -37,6 +38,7 @@ subsets:
   preload:
     subset:
       - loadDiaCatalogs
+      - mpSkyEphemerisQuery
     description: Tasks that can be run before receiving raw images.
   prompt:
     subset:
--- a/pipelines/_ingredients/ApPipe.yaml
+++ b/pipelines/_ingredients/ApPipe.yaml
@@ -67,6 +67,8 @@ tasks:
     class: lsst.pipe.tasks.getRegionTimeFromVisit.GetRegionTimeFromVisitTask
     config:
       connections.coaddName: parameters.coaddName
+  mpSkyEphemerisQuery:
+    class: lsst.ap.association.MPSkyEphemerisQueryTask
   diaPipe:
     class: lsst.ap.association.DiaPipelineTask
     config:
@@ -79,6 +81,7 @@ tasks:
     config:
       connections.science: initial_pvi
       connections.coaddName: parameters.coaddName
+
 subsets:
   apPipe:
     subset:
@@ -92,6 +95,7 @@ subsets:
       - rbClassify
       - transformDiaSrcCat
       - getRegionTimeFromVisit
+      - mpSkyEphemerisQuery      
       - diaPipe
       - analyzeAssocDiaSrcCore
       - analyzeTrailedDiaSrcCore
@@ -104,6 +108,7 @@ subsets:
   preload:
     subset:
       - loadDiaCatalogs
+      - mpSkyEphemerisQuery
     description: Tasks that can be run before receiving raw images.
   prompt:
     subset:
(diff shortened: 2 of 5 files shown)


kfindeisen (changes requested):
Looks ok, but there are a few pieces missing for the pipeline to actually be robust. I'd like to take another look once the contracts are in place, since those tend to be tricky to get right.

Also, please remember to [describe the PR in the title](https://developer.lsst.io/work/flow.html#make-a-pull-request) -- people also come across PRs from the GitHub side, and most can't remember arbitrary 5-digit codes. 🙂

pipelines/_ingredients/ApPipe.yaml
```diff
@@ -191,9 +196,18 @@ contracts:
...
+              diaPipe.connections.ConnectionsClass(config=diaPipe).solarSystemObjectTable.name)
+    msg: "mpSkyEphemerisQuery.ssObjects != diaPipe.solarSystemObjectTable"        
   - contract: diaPipe.connections.ConnectionsClass(config=diaPipe).associatedDiaSources.name ==
               analyzeAssocDiaSrcCore.connections.ConnectionsClass(config=analyzeAssocDiaSrcCore).data.name
     msg: "diaPipe.associatedDiaSources != analyzeAssocDiaSrcCore.data"
   - contract: sampleSpatialMetrics.connections.ConnectionsClass(config=sampleSpatialMetrics).spatiallySampledMetrics.name ==
               diffimTaskPlots.connections.ConnectionsClass(config=diffimTaskPlots).data.name
     msg: "sampleSpatialMetrics.spatiallySampledMetrics != diffimTaskPlots.data"
+
+
```
kfindeisen, line 213: Minor issue, but I'm surprised `git` didn't yell at you over these blank lines.

tests/test_pipelines.py
```diff
@@ -72,7 +72,7 @@ def test_datasets(self):
                     "illumMaskedImage", "deferredChargeCalib",
                     # Everything else
                     "skyMap", "gaia_dr3_20230707", "gaia_dr2_20200414", "ps1_pv3_3pi_20170110",
-                    "goodSeeingCoadd", "pretrainedModelPackage",
+                    "goodSeeingCoadd", "pretrainedModelPackage", "regionTimeInfo"
```
kfindeisen, line 75: `regionTimeInfo` should not be a pipeline input, it's made by `getRegionTimeFromVisit`. Are you saying this test still doesn't pass without this change?

kfindeisen (commented):
I'm a bit confused by the ongoing `regionTimeInfo` input in the tests. Is there something wrong with the pipeline, or is it an oversight?

Unfortunately, we probably can't configure `PipelineStepTester` to reject _extra_ inputs, because that would make refcat handling a lot messier.

kfindeisen:
FTR, I'm about to merge #191, which adds new subsets to the pipeline. @isullivan said he can help take care of the merge conflict.

`mpSkyEphemerisQuery` should end up in both the `apPipe` subset (where it is now) and the new `preload` subset, and these definitions need to be mirrored to both DECam pipelines.

tests/test_pipelines.py
Gerenjie: Yes, when I remove this it fails with 
`tests/test_pipelines.py::PipelineDefintionsTestSuite::test_datasets - AssertionError: Got unexpected pure_inputs: ['regionTimeInfo']`

Gerenjie:
> and these definitions need to be mirrored to both DECam pipelines (EDIT: and also ComCamSim after #193 😭).

I see `ap_pipe/pipelines/DECam` and `ap_pipe/pipelines/ComCamSim` but not a second DECam pipeline, even after rebasing.

tests/test_pipelines.py
kfindeisen: Ok, I'll look into it myself. I didn't see anything obviously wrong on the PR.

kfindeisen:
> I see `ap_pipe/pipelines/DECam` and `ap_pipe/pipelines/ComCamSim` but not a second DECam pipeline, even after rebasing.

You don't have both `ApPipe` and `ApPipeWithFakes`?

tests/test_pipelines.py
kfindeisen: Ah, now I see -- `ApPipe` is fine, it's `Ephemerides.yaml` that's causing the problem. Which is fair, because it's in fact _not_ a self-contained pipeline. Moving `getRegionTimeFromVisit` into `Ephemerides` is not viable, because that task is also a prerequisite for `loadDiaCatalogs`.

Honestly, my preferred solution would be to merge `Ephemerides` into `ApPipe` -- I don't think having a separate file for this task adds anything, and it makes the logic of `ApPipe` itself harder to follow because _both_ its inputs and outputs are actually used by tasks that are directly in `ApPipe`. The alternative is to break up `test_datasets` so that it's actually testing `Ephemerides` separately from `ApPipe` and `ApPipeWithFakes`.

kfindeisen:
Can you please fix the commit history before I do a final review? It may be easiest to do it in 2 passes:

1. Getting rid of the duplicate `main` commits. I'm assuming your intended commit history is:
    - [adding ephemerides.yaml, modifying tests, and adding contracts](https://github.com/lsst/ap_pipe/pull/189/commits/7f89f0466d0ce07c6596cdf421c6fea513221445)
    - [adding ephemerides.yaml, modifying tests, and adding contracts](https://github.com/lsst/ap_pipe/pull/189/commits/17da1fb832e080abaece5591518d5058da2e6295)
    - [Added mpSkyEphemerisQuery to preload subset](https://github.com/lsst/ap_pipe/pull/189/commits/03417d98c8d3456a3a54148ebd796bd848a69872)
    - [adding ephemerides.yaml, modifying tests, and adding contracts](https://github.com/lsst/ap_pipe/pull/189/commits/cf05dba149f9c680cffdb2b069ebfac87b4809c1)
    - [Removed Ephemerides.yaml](https://github.com/lsst/ap_pipe/pull/189/commits/b44a97b66f92e4cdca4712317adb232a65f3d291)
2. Squashing the changes into a coherent history. I think that _all_ of the subsequent commits are of the "should have done this the first time" variety?

Gerenjie:
Tried to rebase this the 'right' way with Ian but it failed, so we decided to save a local copy of the files I'd changed, delete  my local branch, make a new one, and force push just a single commit with all of my changes.

pipelines/DECam/ApPipe.yaml
```diff
@@ -48,6 +50,7 @@ subsets:
       - filterDiaSrcCat
       - rbClassify
       - transformDiaSrcCat
+      - mpSkyEphemerisQuery
```
kfindeisen, line 53: This task should _not_ be in `prompt` (more generally, `preload`, `prompt`, and `afterburner` should be disjoint).

kfindeisen (approved):
Looks good, just one small correction.

PR #190, opened 2024-08-03
DM-31833: Move LoadDiaCatalogs to standalone pipelineTask 

isullivan:


pipelines/_ingredients/ApPipeWithFakes.yaml
```diff
@@ -33,6 +33,10 @@ tasks:
       connections.output_catalog: fakes_initial_pvi_catalog
       connections.injection_prefix: parameters.injection_prefix
       connections.injected_prefix: parameters.injected_prefix
+  loadDiaCatalogs:
+    class: lsst.ap.association.LoadDiaCatalogsTask
+    config:
+      apdb_config_url: parameters.apdb_config
```
kfindeisen, line 39: Please add this to the `ApPipe` subset as well -- that subset is defined to include _everything_.

pipelines/_ingredients/ApPipe.yaml
kfindeisen, line 1: Please add the following contracts:
- outputs of `loadDiaCatalogs` match inputs of `diaPipe`
- `apdb_config_url` of `loadDiaCatalogs` matches `diaPipe`

If you add a `LoadDiaCatalogsConfig.doLoadForcedSources`, please also add a contract that `diaPipe.doRunForcedMeasurement` requires `loadDiaCatalogs.doLoadForcedSources` (that's `(not doRunForcedMeasurement) or doLoadForcedSources`).

kfindeisen (changes requested):
Looks good, but please add contracts for the new task.

pipelines/_ingredients/ApPipe.yaml
```diff
@@ -103,6 +103,8 @@ subsets:
       An alias of ApPipe to use in higher-level pipelines.
 contracts:
   - detectAndMeasure.doSkySources == transformDiaSrcCat.doRemoveSkySources
+  # Do not run forced measurement if the diaForcedSources are not loaded.
+  - loadDiaCatalogs.doLoadForcedSources == diaPipe.doRunForcedMeasurement
```
kfindeisen, line 107: Strictly speaking it's fine if `loadDiaCatalogs.doLoadForcedSources` is set but `diaPipe.doRunForcedMeasurement` is not -- do you prefer the stricter condition?

kfindeisen: Please do add a contract for `apdb_config_url`, just in case someone or something bypasses the pipeline parameter.

kfindeisen (approved):
Looks better, but it's still missing a consistency contract for `apdb_config_url`.

pipelines/_ingredients/ApPipe.yaml
isullivan: My thought for the forced sources was that it was wasteful to load them in `loadDiaCatalogs` if we don't run forced photometry in `diaPipe`. On thinking the logic through more carefully now, the preloaded forced sources are *only* used to package with the alerts, while the forced sources created with `diaPipe.doRunForcedMeasurement` are also written to the APDB. So we could potentially turn loading forced sources off to save time, but still generate new forced sources for the APDB. It is less useful to preload forced sources and not run forced photometry, but I could imagine cases where we would want to do that. Both scenarios are slightly unlikely I think, so I am inclined to either leave the strict condition on the assumption that we would turn both off if we turn off either, or remove the contract entirely.

PR #191, opened 2024-08-19
DM-43418: Divide AP pipeline into preload and prompt subsets

kfindeisen:
This PR divides the AP pipelines into `preload`, `prompt`, and `afterburner` subsets, and ensures the pipeline variants are mutually consistent.

### DIFF
--- a/pipelines/_ingredients/ApPipe.yaml
+++ b/pipelines/_ingredients/ApPipe.yaml
@@ -101,6 +101,35 @@ subsets:
       - initialPviCore
     description: >
       An alias of ApPipe to use in higher-level pipelines.
+  preload:
+    subset:
+      - loadDiaCatalogs
+    description: Tasks that can be run before receiving raw images.
+  prompt:
+    subset:
+      - isr
+      - calibrateImage
+      - retrieveTemplate
+      - subtractImages
+      - detectAndMeasure
+      - filterDiaSrcCat
+      - rbClassify
+      - transformDiaSrcCat
+      - diaPipe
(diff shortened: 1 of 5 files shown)


pipelines/_ingredients/ApPipe.yaml
```diff
@@ -101,6 +101,37 @@ subsets:
...
+      - isr
+      - calibrateImage
+      - retrieveTemplate
+      - subtractImages
+      - detectAndMeasure
+      - filterDiaSrcCat
+      - rbClassify
+      - transformDiaSrcCat
+      - diaPipe
+    description: >
```
isullivan, line 119: Please add a few of the metrics tasks to the `prompt` subset, since they are useful to have in the Prompt Processing Chronograf dashboard:
` - initialPviCore`
` - diffimTaskCore`
` - analyzeAssocDiaSrcCore`
` - analyzeTrailedDiaSrcCore`

isullivan (approved):
I'd like to move a few of the metrics tasks back in to the `prompt` subset, since they should be lightweight and very useful for real-time monitoring.

pipelines/_ingredients/ApPipe.yaml
kfindeisen: To confirm, `injectedMatch` still belongs in the afterburner?

isullivan: No, it had better go in the `prompt` subset. It won't usually matter, but if it does we will want the Fakes metrics before the afterburner runs.

PR #193, opened 2024-08-30
DM-43418: Divide AP pipeline into preload and prompt subsets

kfindeisen:
This PR fixes a bug overlooked in #191, where the ComCamSim specialization of ApPipe is missing the `apPipe` and `prompt` subsets


isullivan (approved):
Looks good.

PR #194, opened 2024-08-31
DM-46050: Use average PSF for packaged alerts

isullivan:
Significantly faster than recalculating the PSF at the location of every source, but not as accurate

### DIFF
--- a/pipelines/_ingredients/ApPipe.yaml
+++ b/pipelines/_ingredients/ApPipe.yaml
@@ -75,6 +75,7 @@ tasks:
       doConfigureApdb: False
       apdb_config_url: parameters.apdb_config
       connections.exposure: initial_pvi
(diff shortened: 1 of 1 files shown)


pipelines/_ingredients/ApPipe.yaml
```diff
@@ -72,6 +72,7 @@ tasks:
     config:
       doConfigureApdb: False
       apdb_config_url: parameters.apdb_config
+      alertPackager.useAveragePsf: True  # Not added to ApPipeWithFakes
```
kfindeisen, line 75: Commit message could be a bit clearer; maybe:
```suggestion
      alertPackager.useAveragePsf: True  # Speed up production processing; don't want as default or in ApPipeWithFakes
```

BrunoSanchez (approved):

PR #192, opened 2024-08-28
DM-39900: Develop improved metrics for completeness based on injected sources

BrunoSanchez:
This PR adds fake source injection matching Signal to Noise (S/N) ratio estimation, and related metrics in analysis tools for completeness as function of this S/N.

PR #195, opened 2024-09-06
DM-39900: DECam pipeline fixup

kfindeisen:
This PR fixes inconsistent subset definitions introduced in #192.


isullivan (approved):
Thank you for catching these!

PR #199, opened 2024-09-27
DM-40855: Add deprecation warning for CreateApFakes task

BrunoSanchez:


python/lsst/ap/pipe/createApFakes.py
```diff
@@ -31,12 +31,18 @@
...
 from lsst.source.injection import generate_injection_catalog
 
+from deprecated.sphinx import deprecated
 
 __all__ = ["CreateRandomApFakesTask",
            "CreateRandomApFakesConfig",
            "CreateRandomApFakesConnections"]
 
 
+@deprecated(
```
isullivan, line 41: No need to add a deprecation warning to the Connections class

isullivan (approved):
Looks good. I believe the warning on the Connections class is redundant, and can be removed.

PR #202, opened 2024-10-15
DM-46356: Update LATISS pipeline to use IsrTaskLSST.

erykoff:


hsinfang:
What collection has the new LATISS calib for recent exposures?
I tried to test this with the collection `LATISS/calib/DM-46359` at `/repo/embargo_old` but it seemed to miss some datasets?

hsinfang (approved):

PR #203, opened 2024-10-21
DM-46105: Add matching for diaSrcs and split metrics between pre & post filtering diaSrcs

BrunoSanchez:
Modification of the matching task to include flag information and diaSources-assocDiaSources.

### DIFF
--- a/python/lsst/ap/pipe/matchSourceInjected.py
+++ b/python/lsst/ap/pipe/matchSourceInjected.py
@@ -56,9 +56,15 @@ class MatchInitialPVIInjectedConnections(
         storageClass="ExposureF",
         dimensions=("instrument", "visit", "detector"),
     )
-    associatedDiaSources = connTypes.Input(
-        doc="A DiaSource catalog to match against fakeCat. Assumed "
-            "to be SDMified.",
+    diaSources = connTypes.Input(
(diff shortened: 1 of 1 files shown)


python/lsst/ap/pipe/matchSourceInjected.py
```diff
@@ -202,7 +215,7 @@ def _estimateFakesSNR(self, injectedCat, diffIm):
 
         # Add the forced measurement columns to the input catalog
         for column in forcedSources_table.columns:
-            if "Flux" in column:
+            if "Flux" in column or "flag" in column:
```
isullivan, line 218: Can we get these columns from the schema set by the forced source task instead? I am concerned that this might copy too many columns, and might miss some.

isullivan (approved):
Looks good! A few minor points for you to consider.

python/lsst/ap/pipe/matchSourceInjected.py
BrunoSanchez: I don't know exactly how to make this be set by the forced source task instead.
Would this be just by re-defining the schema?

isullivan: I misunderstood: I thought this was renaming columns, but I see it is choosing columns to copy over to the injected catalog. That seems fine!

PR #207, opened 2024-11-20
DM-47620 Adding single-frame fallback pipelines

Gerenjie:
My understanding of why I was instructed to make these changes in the particular structure I did: 

As-is, SingleFrame.yaml exists in `prompt_processing` but nowhere else. We want a system in which `ap_pipe` (and `ci_pipe` and `drp_pipe`) sit on top of everything else and describe pipeline functionality via yamls. All `prompt_processing` pipeline yamls should reference and modify counterparts in `ap_pipe`. At the moment, SingleFrame.yaml is a standalone `prompt_processing` pipeline, so in addition to adding single-frame association to it, I was asked to also move it into `ap_pipe` in a traditional LSSTComCam -> _ingredients reference chain.

### DIFF
--- a/pipelines/_ingredients/SingleFrame.yaml
+++ b/pipelines/_ingredients/SingleFrame.yaml
@@ -1,0 +1,19 @@
+description: Single-frame pipeline for the case in which no templates exist
+imports:
+  - location: $AP_PIPE_DIR/pipelines/_ingredients/ApPipe.yaml
+    include:
+      - processCcd
+      - initialPviCore
+      - getRegionTimeFromVisit
+      - mpSkyEphemerisQuery
+tasks:
+  ssSingleFrameAssociation:
+    class: lsst.ap.association.SsSingleFrameAssociationTask 
+subsets:
+  processCcd:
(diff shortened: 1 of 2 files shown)


mrawls (approved):
This feels very convoluted, and will be simpler if/when we get rid of ProcessCcd.yaml as its own standalone thing. But you say ci_ap passes, and nothing seems out of order, so go ahead.

pipelines/_ingredients/SingleFrame.yaml
```diff
@@ -0,0 +1,19 @@
...
+  ssSingleFrameAssociation:
+    class: lsst.ap.association.SsSingleFrameAssociationTask 
+subsets:
+  processCcd:
+    subset:
+      - isr
+      - calibrateImage
+      - ssSingleFrameAssociation
+    description: >
+      An alias of SingleFrame to use in higher-level pipelines.
```
mrawls, line 19: Please edit this description because calling processCcd an alias of singleFrame is quite confusing.

PR #208, opened 2024-11-21
DM-47737 Adding SingleFrame for all instruments

Gerenjie:
Add SingleFrame pipeline for any instruments which don't already have it.

### DIFF
--- a/pipelines/DECam/SingleFrame.yaml
+++ b/pipelines/DECam/SingleFrame.yaml
@@ -1,0 +1,22 @@
+description: >-
+  Single-frame pipeline for the case in which
+  no templates exist, specialized for DECam
+instrument: lsst.obs.decam.DarkEnergyCamera
+imports:
+  - location: $AP_PIPE_DIR/pipelines/DECam/ProcessCcd.yaml
+    include: 
+      - processCcd
+  - location: $AP_PIPE_DIR/pipelines/_ingredients/SingleFrame.yaml
+    include:
+      - getRegionTimeFromVisit
+      - mpSkyEphemerisQuery
+      - ssSingleFrameAssociation
+      - initialPviCore
+subsets:
+  processCcd:
+    subset:
+      - isr
+      - calibrateImage
+      - ssSingleFrameAssociation
+    description: >
+      Includes processCcd plus ssSingleFrameAssociation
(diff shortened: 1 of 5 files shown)


pipelines/DECam/SingleFrame.yaml
```diff
@@ -0,0 +1,23 @@
...
+imports:
+  - location: $AP_PIPE_DIR/pipelines/DECam/ProcessCcd.yaml
+    include: 
+      - processCcd
+  - location: $AP_PIPE_DIR/pipelines/_ingredients/SingleFrame.yaml
+    include:
+      - getRegionTimeFromVisit
+      - mpSkyEphemerisQuery
+      - ssSingleFrameAssociation
+      - initialPviCore
```
kfindeisen, line 14: The following might be more robust to pipeline changes, but hopefully it becomes irrelevant when we remove `ProcessCcd.yaml`:
```suggestion
    exclude:
      - isr
      - calibrateImage
```

pipelines/DECam/SingleFrame.yaml
```diff
@@ -0,0 +1,23 @@
...
+subsets:
+  processCcd:
+    subset:
+      - isr
+      - calibrateImage
+      - ssSingleFrameAssociation
+      - initialPviCore
+    description: >
+      This adds ssSingleFrameAssociation to processCcd
```
kfindeisen, line 23: The description is what people will see when inspecting the pipeline with `pipetask`, so it should make sense in a standalone context.

pipelines/DECam/SingleFrame.yaml
```diff
@@ -0,0 +1,23 @@
...
+      - mpSkyEphemerisQuery
+      - ssSingleFrameAssociation
+      - initialPviCore
+subsets:
+  processCcd:
+    subset:
+      - isr
+      - calibrateImage
+      - ssSingleFrameAssociation
+      - initialPviCore
```
kfindeisen, line 21: I'd leave out `initialPviCore`, for consistency with the definition in `_ingredients` (though I agree we probably will want to add it everywhere in the future).

kfindeisen (approved):
Looks good, minor suggestions.

PR #210, opened 2024-11-22
DM-35504: Unify ap_pipe configs

kfindeisen:
This PR removes the last traces of our `obs_*` hacks and centralizes the remaining config (actually parameter) overrides in `_ingredients`. It also addresses a deprecation warning in the unit tests.


mrawls (approved):
Thank you for cleaning this up!

PR #211, opened 2024-11-26
DM-47800: Set limit on number of refcat sources for astrometric fit

isullivan:


kfindeisen (approved):
I'm a bit surprised this is all that's needed.

isullivan:
Closing without merging, since the default was moved to a lower level.

PR #212, opened 2024-12-02
DM-47892 : adding associationCore to all ApPipe variants

Gerenjie:


pipelines/_ingredients/ApPipe.yaml
```diff
@@ -120,6 +121,7 @@ subsets:
       - analyzeAssocDiaSrcCore
       - analyzeTrailedDiaSrcCore
       - diffimTaskCore
+      - associationCore
```
kfindeisen, line 124: Can the management confirm this is something we want in `prompt` and not `afterburner`?

isullivan: Yes, this is lightweight and is intended to be run in Prompt Processing so that these metrics can show up in Chronograf.

PR #213, opened 2024-12-03
DM-47899: Turn on rbClassify for LsstComCam and LsstComCamSim

isullivan:
Remove the exclude, and the subset definitions since those can now be inherited directly from `_ingredients/ApPipe.yaml`


kfindeisen (approved):
Looks good. Closes lsst-dm/prompt_processing#236 (though I'm not sure GitHub's auto-close works in review comments).

kfindeisen:
For the record, auto-close did not work. 😞 I'll keep it open until we confirm the new daily is OK.

PR #215, opened 2024-12-11
DM-46098: Add detectionTaskCore to all pipelines

isullivan:


bps/clustering/clustering_ApPipe.yaml
```diff
@@ -23,7 +23,7 @@ cluster:
     dimensions: visit,detector
     equalDimensions: visit:exposure
   diffim:
-    pipetasks: retrieveTemplate,subtractImages,detectAndMeasure,diffimTaskCore,filterDiaSrcCat,sampleSpatialMetrics,diffimTaskPlots,rbClassify,transformDiaSrcCat
+    pipetasks: retrieveTemplate,subtractImages,detectAndMeasure,diffimTaskCore,detectionTaskCore,filterDiaSrcCat,sampleSpatialMetrics,diffimTaskPlots,rbClassify,transformDiaSrcCat
```
kfindeisen, line 26: Question for #216: is there a way to inspect the clustering config from Python, and any hard requirements for what should be in it (I assume there's some reason the list of tasks needs to be manually specified)?

kherner (approved):

PR #218, opened 2025-01-07
DM-48122: Remove ProcessCcd.yaml from ap_pipe

kfindeisen:
This PR removes `ProcessCcd.yaml` and `RunIsrWithCrosstalk.yaml`, moving their contents into the main pipelines and a config file, respectively. This change removes all pipeline splicing from `ap_pipe` (though we still have some in `ap_verify`).

With the removal of `ProcessCcd.yaml`, this PR also deprecates the `processCcd` subset, replacing it with `apPipeSingleFrame`.


pipelines/_ingredients/ApPipeWithFakes.yaml
```diff
@@ -119,11 +124,21 @@ tasks:
...
 subsets:
+  apPipeSingleFrame:
+    subset:
+      - isr
+      - calibrateImage
+      - initialPviCore
+    description: >-
+      The prompt ApPipe tasks that make up single-frame processing. Not to be confused with the
+      SingleFrame.yaml pipeline, which is designed to be run instead of ApPipeWithFakes.yaml and
+      includes some preload tasks and some non-ApPipe tasks.
```
mrawls, line 135: Suggested reword: "Not to be confused with `SingleFrame.yaml`, which includes `apPipeSingleFrame`, and is designed as an alternative to `ApPipeWithFakes.yaml` in the situation where templates for difference imaging are not available.

pipelines/_ingredients/SingleFrame.yaml
```diff
@@ -2,7 +2,7 @@ description: Single-frame pipeline for the case in which no templates exist
 imports:
   - location: $AP_PIPE_DIR/pipelines/_ingredients/ApPipe.yaml
     include:
-      - processCcd
+      - apPipeSingleFrame
       - initialPviCore
```
mrawls, line 6: The new `apPipeSingleFrame` includes `initialPviCore`, which suggests you probably don't need it listed here too.

mrawls (approved):
DECam is so much nicer now! As far as I can tell, this looks good, minor comments notwithstanding, and I appreciate you running the CI data sets to verify (ha) they all still work.

pipelines/_ingredients/ApPipeWithFakes.yaml
kfindeisen: I'm not sure I agree:
* Saying that `SingleFrame.yaml` includes `apPipeSingleFrame`, while technically true, makes them sound related, which is exactly the confusion I wanted to avoid.
* I don't want to say that `SingleFrame.yaml` is for "the situation where templates are not available", and objected to that wording when it was put in the file. Describing components as being for some assumed context (in this case, for a package that `ap_pipe` isn't supposed to depend on) leads to nonmodularity and future bugs. As far as I'm concerned, `SingleFrame` is a standalone pipeline that's not a subset of `ApPipe`, no more, no less.

mrawls: OK fair, I appreciate you want to avoid conflating "what it is" and "what we often use it for" in docs like this.

How about, "Not to be confused with the `SingleFrame.yaml` pipeline, which does more than just ApPipe single frame processing, and is designed as an alternative to `ApPipeWithFakes.yaml`." ?

PR #223, opened 2025-02-19
DM-48896 : Create yaml for ComCam

BrunoSanchez:
This has evolved and now it is scoped for replacing all pipeline creation

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -1,9 +1,8 @@
 # -*- python -*-
 import os
 
-from lsst.sconsUtils import scripts
+from lsst.sconsUtils import scripts, targets
 from lsst.sconsUtils.state import env
-from lsst.sconsUtils import targets
 from lsst.sconsUtils.utils import libraryLoaderEnvironment
 from SCons.Script import Default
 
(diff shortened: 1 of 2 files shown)


SConstruct
```diff
@@ -1,3 +1,59 @@
...
+subset_names = ["apPipe", "prompt"]
+
+ingredients_ap_pipe_with_fakes = env.Command(
+    target=ap_pipe_with_fakes_path,
+    source=os.path.join(PKG_ROOT, "pipelines", "_ingredients", "ApPipe.yaml"),
+    action=" ".join(
+        [
+            libraryLoaderEnvironment(),
+            f"make_injection_pipeline  -t preliminary_visit_image -r $SOURCE -f $TARGET "
+            f"-a {additional_fakes_tasks} -s "+" -s ".join(subset_names),
```
leeskelvin, line 37: Should `-s` be used twice here?

leeskelvin: And if `subset_names` is length zero, this still prints a single `-s` then? Is that an issue?

BrunoSanchez: Adopted! Thanks

PR #225, opened 2025-03-27
DM-49740: Start LSSTCam pipeline in ap_pipe

hsinfang:


pipelines/LSSTCam/ApPipe.yaml
```diff
@@ -0,0 +1,7 @@
+description: >-
+  AP pipeline specialized for LSSTCam Prompt Processing. This
+  file is imported in prompt_processing/pipelines/LSSTCam/ApPipe.yaml
+  which is used in realtime Prompt Processing production.
```
kfindeisen, line 4: I realize this is a pre-existing issue, but I find both these descriptions a little problematic. The AP pipeline is simply the AP pipeline; while we do use it in PP, `ap_pipe` shouldn't assume or depend on that fact. Maybe:
```suggestion
  AP pipeline specialized for LSSTCam.
```

pipelines/LSSTCam/SingleFrame.yaml
```diff
@@ -0,0 +1,6 @@
+description: >-
+  Single-frame pipeline for the case in which
+  no templates exist, specialized for LSSTCam
```
kfindeisen, line 3: Again, no need to assume _why_ the user might want to run it (it's not like having templates will somehow invalidate the pipeline!):
```suggestion
  Standalone single-frame pipeline, specialized for LSSTCam.
```

kfindeisen (approved):
I'm a bit worried about the proliferation of docs that make specific assumptions about a package that is above `ap_pipe` in the dependency tree. Would it be possible to describe them a bit more generically?

PR #235, opened 2025-07-22
DM-50987: Integrate sattle with Prompt Processing

ebellm:


pipelines/_ingredients/ApPipe.yaml
```diff
@@ -248,6 +248,7 @@ contracts:
   - detectAndMeasureDiaSource.doSkySources == standardizeDiaSource.doRemoveSkySources
   # Both loadDiaCatalogs and associateApdb connect to the APDB, so make sure they use the same configuration
   - loadDiaCatalogs.apdb_config_url == associateApdb.apdb_config_url
+  - calibrateImage.run_sattle == detectAndMeasureDiaSource.run_sattle
```
kfindeisen, line 251: Maybe a comment to the effect that `calibrateImage.run_sattle` isn't _strictly_ necessary but is an important optimization?

kfindeisen (approved):
I'd prefer a pipeline parameter to make it easier/safer to keep the two options in sync, but I suppose we're confident that only PP will override them?

PR #238, opened 2025-08-13
DM-24858: Add parameter for release_id with default 0

isullivan:


pipelines/_ingredients/ApPipe.yaml
```diff
@@ -19,6 +19,8 @@ parameters:
   coaddName: goodSeeing
   # APDB config file must be user-specified
   apdb_config:
+  # Use release_id=0 for development, 1 for production, and 2 or 3 for reprocessing production data
+  release_id: 1
```
kfindeisen, line 23: If production is the default, then Prompt Processing doesn't need to do anything and DM-52170 is a no-op. However, _do_ we want production to be the default? Dev environments tend to be less tightly controlled, so it's more likely that someone would forget to override there.

kfindeisen: I recommend overriding `release_id=0` in `ap_verify` regardless of the above.

pipelines/_ingredients/ApPipe.yaml
```diff
@@ -158,6 +162,7 @@ tasks:
       connections.diaObjects: "{fakesType}dia_object_apdb"
       connections.newDiaSources: "{fakesType}new_dia_source"
       connections.marginalDiaSources: "{fakesType}marginal_new_dia_source"
+      idGenerator.release_id: parameters.release_id
```
kfindeisen, line 165: To confirm, the release ID is needed in `calibrateImage`, `detectAndMeasureDiaSource`, and `associateApdb`? Why would the last need to generate IDs?

I can't think of any tasks that might be missing, but the pipeline isn't my area of expertise.

kfindeisen (approved):
Looks good, no obvious gaps. My only serious question is about the default handling.

pipelines/_ingredients/ApPipe.yaml
isullivan: Thank you for catching this! I meant to use `release_id: 0`, but accidentally committed an edit I used for testing. I'll change that, and set the parameter explicitly in `ap_verify` as you suggest.

pipelines/_ingredients/ApPipeWithFakes.yaml
```diff
@@ -18,6 +18,8 @@ parameters:
   fakesType: 'fakes_'
   injected_prefix: 'fakes_'
   injection_prefix: 'injection_'
+  # Use release_id=0 for development, 1 for production, and 2 or 3 for reprocessing production data
+  release_id: 0
```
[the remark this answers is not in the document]
ebellm: We don't yet have any "production" fakes runs, so I think 0 is fine.  Whether we want to reserve a whole release id for future fakes processing would warrant some discussion.

PR #240, opened 2025-08-24
DM-52288: Filter low-reliability DIASources for first alerts

ebellm:


kfindeisen (changes requested):
The current changes touch only the precursor instruments, as opposed to LSSTCam.

PR #266, opened 2026-04-14
DM-54643: Fix optional pyproj import in test_pipelines.py

roceb:
Guard the pyproj import with a try/except block so that the test
module can be loaded even when pyproj is not installed. The import
was only needed to work around a file handle leakage issue and is
not required for the tests themselves to function.

### DIFF
--- a/tests/test_pipelines.py
+++ b/tests/test_pipelines.py
@@ -24,7 +24,11 @@ import tempfile
 import unittest
 
 # need to import pyproj to prevent file handle leakage
-import pyproj  # noqa: F401
+# TODO: Remove import after completing DM-54643. Use DM-54656
+try:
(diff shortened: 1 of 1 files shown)


tests/test_pipelines.py
```diff
@@ -24,7 +24,11 @@
 import unittest
 
 # need to import pyproj to prevent file handle leakage
-import pyproj  # noqa: F401
+# TODO: Remove import after completing DM-54643.
```
timj: I think we also want to say which ticket the work is going to be done on after DM-54643 is completed otherwise someone is bound to make a whole new ticket for it.

timj (approved):
Looks okay. One comment. Assume I am approving the other 3 with the same comment.

PR #269, opened 2026-04-20
DM-54625: Add preconvolution pipelines

isullivan:


pipelines/_ingredients/ApPipeWithPreconvolution.yaml
```diff
@@ -0,0 +1,107 @@
...
+    class: lsst.ip.diffim.subtractImages.AlardLuptonPreconvolveSubtractTask
+    config:
+      connections.coaddName: parameters.coaddName
+      connections.science: preliminary_visit_image
+      connections.sources: single_visit_star_footprints
+      connections.template: template_detector
+      connections.psfMatchingKernel: score_kernel
+      connections.kernelSources: score_kernel_sources
+      connections.scoreExposure: score_image
+      doApplyExternalCalibrations: False
```
kfindeisen, line 22: Do any of the `score_*` datasets need to be persisted from Prompt Processing?

pipelines/_ingredients/ApPipeWithPreconvolution.yaml
```diff
@@ -0,0 +1,107 @@
...
+subsets:
+  apPipeSingleFrame:
+    subset:
+      - isr
+      - calibrateImage
+      - analyzePreliminarySummaryStats
+    description: >-
+      The prompt ApPipe tasks that make up single-frame processing. Not to be confused with the
+      SingleFrame.yaml pipeline, which does more than just ApPipe single frame processing, and
+      is designed as a standalone alternative to ApPipe.yaml.
```
kfindeisen: There's no need to redefine subsets if they're not changing -- all it does is introduce the risk of code drift.

I think in this case that means you only need to change `apPipe` and `prompt`?

pipelines/DECam/ApPipeWithPreconvolution.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+instrument: lsst.obs.decam.DarkEnergyCamera
+imports:
+  - location: $AP_PIPE_DIR/pipelines/_ingredients/ApPipeWithPreconvolution.yaml
+tasks:
+  isr:
+    class: lsst.ip.isr.IsrTask
+    config:
+      file: $AP_PIPE_DIR/config/DECam/runIsrWithCrosstalk.py
+contracts:
```
kfindeisen, line 20: These should go in `_ingredients` -- the structure/integrity of the pipeline doesn't depend on the instrument.

kfindeisen: I don't see any links (forward or backward) for `subtractImagesScore`?

pipelines/DECam/ApPipeWithPreconvolution.yaml
```diff
@@ -0,0 +1,35 @@
...
+# (3) Run this pipeline, setting the apdb_config parameter to point to the new file
+
+instrument: lsst.obs.decam.DarkEnergyCamera
+imports:
+  - location: $AP_PIPE_DIR/pipelines/_ingredients/ApPipeWithPreconvolution.yaml
+tasks:
+  isr:
+    class: lsst.ip.isr.IsrTask
+    config:
+      file: $AP_PIPE_DIR/config/DECam/runIsrWithCrosstalk.py
```
kfindeisen, line 17: This is incorrect, because `ApPipeWithPreconvolution` is based on `ApPipeWithIsrTaskLSST`, which isn't applicable to DECam. You could get around the problem by splicing `_ingredients/ApPipeWithPreconvolution` and `DECam/ApPipe`, but that would have the same contract/subset problems we get in `ap_verify`.

I suppose one question to ask is whether we actually *need* `ApPipeWithPreconvolution` for non-LSST cameras.

kfindeisen (changes requested):
I'm pretty sure the DECam/HSC versions of the new pipeline are broken -- they use the wrong ISR task. Unfortunately, I don't see a good fix, short of having separate LSST and non-LSST versions in `_ingredients`.

The other problem is that the contracts have been deferred to the instrument-specific tasks, and that means the ones for `subtractImagesScore` have slipped through the cracks.

pipelines/DECam/ApPipeWithPreconvolution.yaml
isullivan: This gets at the central problem I was trying to work around, but maybe it is too ugly of a kludge. I only wanted to make a single `ApPipeWithPreconvolution` and not duplicate it for `ApPipeWithIsrTaskLSST` and `ApPipe`, so for DECam and HSC I still inherit from `ApPipeWithIsrTaskLSST` but override the `isr` definition back to `lsst.ip.isr.IsrTask`. This is clearly misleading, but it does work. I would rather not do the splicing approach because of the contract and subset problems you mention, but dropping `ApPipeWithPreconvolution` for non-LSST cameras seems like a reasonable option.
I am willing to do the splicing approach if you think that would be best, drop `ApPipeWithPreconvolution` for the other cameras, or leave it as-is but add comments explaining what is going on and why. Your call.

pipelines/_ingredients/ApPipeWithPreconvolution.yaml
isullivan: None of them are required and we are short on disk space, so I think we won't plan to persist them from Prompt Processing.

pipelines/DECam/ApPipeWithPreconvolution.yaml
isullivan: > These should go in `_ingredients` -- the structure/integrity of the pipeline doesn't depend on the instrument.

I was copying the contracts specified in `pipelines/DECam/ApPipe.yaml`. Should that not include contracts?

isullivan: > I don't see any links (forward or backward) for `subtractImagesScore`?

Sorry, what do you mean by forward or backward links?

kfindeisen: Re: `ApPipe.yaml`, I take it that's #228? So be it, though the comment could be a bit clearer.

I meant that there are no contracts verifying that the inputs of `subtractImagesScore` match the outputs of the task(s) that create them, nor that the outputs match the inputs of downstream tasks (in particular, `detectAndMeasureDiaSource`). Basically, what the bulk of the contracts in `_ingredients/ApPipe.yaml` are for.

pipelines/DECam/ApPipeWithPreconvolution.yaml
kfindeisen: Oops, you're right, I don't know how I missed that. 🤦 Yes, I agree that overriding the ISR definition is probably best (`IsrTaskLSST`'s connections seem to be a bit unstable, but hopefully `IsrTask` isn't changing anymore?).

I think highlighting the class change with a comment might be good enough. You might also add a unit test to guard against code drift (maybe that `<instrument>/ApPipe` and `<instrument>/ApPipeWithPreconvolution` have identical ISR task definitions?)

pipelines/DECam/ApPipeWithPreconvolution.yaml
isullivan: Yes, the missing contracts were an oversight. I've added some.

pipelines/DECam/ApPipeWithPreconvolution.yaml
isullivan: Good idea. I have added a unit test that verifies that the ISR task configs are the same between ApPipe and ApPipeWithPreconvolution versions of the pipeline.

isullivan:
@kfindeisen I believe this PR is ready for re-review.

pipelines/DECam/ApPipeWithPreconvolution.yaml
```diff
@@ -0,0 +1,37 @@
...
+    msg: "calibrateImage.exposure != computeReliability.science"
+  - contract: calibrateImage.connections.ConnectionsClass(config=calibrateImage).exposure.name ==
+              makeSampledImageSubtractionMetrics.connections.ConnectionsClass(config=makeSampledImageSubtractionMetrics).science.name
+    msg: "calibrateImage.exposure != makeSampledImageSubtractionMetrics.science"
+  - contract: calibrateImage.connections.ConnectionsClass(config=calibrateImage).exposure.name + ".summaryStats" ==
+              analyzePreliminarySummaryStats.connections.ConnectionsClass(config=analyzePreliminarySummaryStats).data.name
+    msg: "calibrateImage.exposure != analyzePreliminarySummaryStats.data"
+  - contract: calibrateImage.connections.ConnectionsClass(config=calibrateImage).stars_footprints.name ==
+              getRegionTimeFromVisit.connections.ConnectionsClass(config=getRegionTimeFromVisit).dummy_visit.name
+    msg: "calibrateImage.stars_footprints != getRegionTimeFromVisit.dummy_visit"
```
kfindeisen, line 39: Missing the contracts checking consistency of `calibrateImage` with `subtractImagesScore` -- or are the checks vs. `subtractImages` supposed to take care of that?

pipelines/HSC/ApPipeWithPreconvolution.yaml
```diff
@@ -0,0 +1,37 @@
...
+#     apdb-cli create-sql "sqlite:////project/user/association.db" apdb_config.yaml
+#     apdb-cli metadata set apdb_config.yaml instrument HSC
+# (2) Run this pipeline, setting the apdb_config parameter to point to the new file
+
+instrument: lsst.obs.subaru.HyperSuprimeCam
+imports:
+  - location: $AP_PIPE_DIR/pipelines/_ingredients/ApPipeWithPreconvolution.yaml
+tasks:
+  # Override the class for `isr`, which is set to `IsrTaskLSST` in the base
+  # pipeline in `_ingredients`
```
kfindeisen, line 12: This comment seems to be missing from the DECam version.

tests/test_pipelines.py
```diff
@@ -180,6 +180,47 @@ def test_ap_pipe_subsets(self):
...
+        ]
+        # Sanity-check that this test actually has cameras to compare.
+        self.assertGreater(len(files), 0,
+                           msg="No camera-specific ApPipeWithPreconvolution.yaml files found.")
+
+        for precon_file in files:
+            with self.subTest(file=str(precon_file)):
+                base_file = precon_file.dirname().join("ApPipe.yaml")
+                self.assertTrue(base_file.exists(),
+                                msg=f"Expected sibling ApPipe.yaml next to {precon_file}.")
```
kfindeisen: Might be better to report `base_file` directly, just in case there's a weird bug in defining it.

tests/test_pipelines.py
```diff
@@ -180,6 +180,47 @@ def test_ap_pipe_subsets(self):
...
+                base_isr = base.to_graph().tasks["isr"]
+
+                self.assertEqual(precon_isr.task_class_name, base_isr.task_class_name,
+                                 msg=f"isr task class differs between ApPipe.yaml and "
+                                     f"ApPipeWithPreconvolution.yaml in {precon_file.dirname()}.")
+                self.assertTrue(
+                    base_isr.config.compare(precon_isr.config, shortcut=False),
+                    msg=f"isr task config differs between ApPipe.yaml and "
+                        f"ApPipeWithPreconvolution.yaml in {precon_file.dirname()}."
+                )
```
kfindeisen, line 225: [Ugh](https://pipelines.lsst.io/v/daily/api/lsst.pipe.base.pipeline_graph.TaskNode.html#notes):
> Task nodes are intentionally not equality comparable, since there are many different (and useful) ways to compare these objects with no clear winner as the most obvious behavior.

Maybe add a comment that you can't just do `assertEqual(precon_isr, base_isr)`?

kfindeisen, line 225: For this, you could probably get away with `assertEqual` -- since the configs are being read directly from files, any floating-point handling should be deterministic.

kfindeisen (approved):
Looks good, assuming GitHub posts all my comments (I got some weird errors trying to record them).

pipelines/DECam/ApPipeWithPreconvolution.yaml
isullivan: No harm in adding those contracts as well.

PR #272, opened 2026-05-13
DM-54077: Fix missing testing for dynamically generated Fakes pipelines

BrunoSanchez:


tests/test_pipelines.py
```diff
@@ -70,19 +70,19 @@ def test_graph_build(self):
...
         for file in files:
             if "QuickTemplate" in file.path:
                 # Our QuickTemplate definition cannot be tested here because it
                 # depends on drp_tasks, which we cannot make a dependency here.
                 continue
+            if "injection/" in file.path:
+                # The source-injection post-processing ingredient is a partial
+                # pipeline merged into full AP pipelines at build time;
+                # it is validated separately by test_injection_ingredient.
+                continue
```
kfindeisen, line 85: I think these diffs (and the matching ones below) would have made sense as their own commit, since they're self-contained (i.e., cleaning up how `injection/` is excluded) and distinct from the addition of `test_injection_ingredient`.

tests/test_pipelines.py
```diff
@@ -96,7 +96,11 @@ def test_datasets(self):
                     "skyMap", "gaia_dr3_20230707", "gaia_dr2_20200414", "ps1_pv3_3pi_20170110",
                     "template_coadd", "pretrainedModelPackage", "dia_source_apdb"
                 }
-                if "WithFakes" in file.path:
+                # Detect source-injection pipelines by task label rather than
+                # relying on filename conventions.
+                pipeline_check = lsst.pipe.base.Pipeline.from_uri(file)
+                pipeline_check.addConfigOverride("parameters", "apdb_config", "some/file/path.yaml")
```
kfindeisen: `pipeline_check` seems like the wrong name for what this variable is doing (to me, it sounds like either a function for checking the pipeline, or the result of such a check). Maybe `temp_pipeline`? Or factor it into a `get_task_labels`, then you don't have to worry about getting a unique name?

tests/test_pipelines.py
```diff
@@ -149,18 +153,18 @@ def test_whole_subset(self):
...
         # getRegionTimeFromVisit is part of no subset besides apPipe. This is a
         # very deliberate exception; see RFC-997.
         no_subset_wanted = {"getRegionTimeFromVisit"}
 
         for file in files:
+            if "injection/" in file.path:
+                # Source-injection post-processing ingredients are partial
+                # pipelines; they don't satisfy the full ApPipe subset
+                # convention independently.
+                continue
```
kfindeisen, line 163: This is kind of unfortunate effect of naming the pipeline `ApPipePostInjectedTasks`, when the tests assume that a pipeline starting with `ApPipe` is *some* variant of the AP pipeline. I know you said "no naming conventions" above, but this one seems intuitive enough to be worth keeping.

If not, then I suggest simplifying the comment to "ApPipePostInjectedTasks is not actually an AP pipeline", since that's the essential issue.

tests/test_pipelines.py
```diff
@@ -180,6 +184,57 @@ def test_ap_pipe_subsets(self):
...
+        """Test the source-injection post-processing ingredient pipeline.
+
+        ApPipePostInjectedTasks is a partial pipeline merged into full AP
+        pipelines at build time by make_injection_pipeline. This test
+        validates that it can build a graph and contains the expected tasks.
+        """
+        ingredient = (
+            self.path
+            .join("_ingredients/", forceDirectory=True)
+            .join("injection/", forceDirectory=True)
```
kfindeisen: I suspect the "/" and `forceDirectory=True` are both redundant (though I'm surprised you really do need to chain multiple `join` calls like this).

tests/test_pipelines.py
```diff
@@ -180,6 +184,57 @@ def test_ap_pipe_subsets(self):
...
+        generation occurred before pipeline tests run, and that the generated
+        pipeline includes the source-injection task. If this test is skipped,
+        run 'scons' in the ap_pipe root directory first.
+        """
+        generated = self.path.join("_ingredients/ApPipeWithFakes.yaml")
+        if not generated.exists():
+            self.skipTest(
+                f"{generated} has not been generated yet. "
+                "Run 'scons' in the ap_pipe root directory to generate it."
+            )
```
kfindeisen: Should this really be `skipTest` instead of `fail`? People tend to ignore skipped tests, but if the generated pipeline doesn't exist, then the results of `test_datasets` and `test_whole_subset` are invalid (or, at least, incomplete).

kfindeisen (approved):
Looks good, biggest concern is the `generated_pipeline_readiness` test -- I think it should be stricter, to alert the user that the test assumptions haven't been met.

tests/test_pipelines.py
BrunoSanchez: I agree it must be fail then, I tested this and I think it happens after generation (at least all the times I tried). I will make the change.

tests/test_pipelines.py
BrunoSanchez: I think we can rename the file altogether, there's no problem with that.

tests/test_pipelines.py
```diff
@@ -180,6 +180,57 @@ def test_ap_pipe_subsets(self):
...
+        """Test the source-injection post-processing ingredient pipeline.
+
+        PostInjectedTasksApPipe is a partial pipeline merged into full AP
+        pipelines at build time by make_injection_pipeline. This test
+        validates that it can build a graph and contains the expected tasks.
+        """
+        ingredient = (
+            self.path
+            .join("_ingredients/injection/PostInjectedTasksApPipe.yaml")
+        )
```
kfindeisen: I think the original chain of `.join` was safer than having hardcoded slashes -- my point was that you didn't need to meticulously declare that all the intermediates are directories.
```suggestion
        ingredient = self.path.join("_ingredients").join("injection").join("ApPipePostInjectedTasks.yaml")
```

kfindeisen (approved):

andy-slac:
@BrunoSanchez, it looks like this PR has broken Jenkins. I think that the test that you added looks for a file `pipelines/_ingredients/injection/ApPipePostInjectedTasks.yaml` which does not exist. Did you forget to commit that file?

kfindeisen:
Sorry, ignore my last post -- it's an incomplete rename, fix should be to change all references to `ApPipePostInjectedTasks` with `PostInjectedTasksApPipe`.

andy-slac:
I'll make a ticket and work on a quick fix.

PR #275, opened 2026-05-28
DM-50420: Inject variable sources into the template and science images

BrunoSanchez:
Create the fake injection pipeline tools for injection on template images.

### DIFF
--- a/doc/lsst.ap.pipe/ap-pipe-pipeline-with-fakes.rst
+++ b/doc/lsst.ap.pipe/ap-pipe-pipeline-with-fakes.rst
@@ -61,3 +61,49 @@ The key additional outputs include:
 * ``fakes_preliminary_visit_image_catalog``: Catalog of injected source properties
 * ``fakes_goodSeeingDiff_matchDiaSrc``: Matches between injected and detected sources
 * ``fakes_goodSeeingDiff_matchAssocDiaSrc``: Matches for successfully associated sources
+
+
+Creating an Injection Catalog
+-----------------------------
+To create an injection catalog for use with ``ApPipeWithFakes``, use the ``createInjectionCatalog`` command from the ``lsst.source.injection`` package * :ref:`source_injection <lsst.source.injection-ref-generate>`. This command generates a catalog of synthetic sources based on specified parameters such as magnitude distribution, spatial distribution, and source types.
+Alternatively, you can use the ``createApFakes`` task from the ``lsst.ap.pipe`` package, which is specifically designed to create injection catalogs tailored for the AP pipeline.
+We also provide a pipeline yaml file that is able to create an injection catalog with the same configuration as the one used in the ``ApPipeWithFakes`` pipeline, which can be found at ``pipelines/CreateInjectionCatalogs.yaml``. This pipeline can be run using the same command as for running the AP pipelines, but with the above yaml file as input and specifying an output collection to ingest the generated injection catalog into.
+
+.. code-block:: bash
+
+    pipetask run  \
+        -i u/user/input_collection \
+        -b /path/to/repo/butler.yaml \
+        -o fake_catalogs_collection \
+        -p $AP_PIPE_DIR/pipelines/CreateInjectionCatalogs.yaml \
+        -d "instrument='LSSTComCam' AND skymap='lsst_cells_v2' AND visit=2024111100094 AND detector IN (0,3)"
+
+
+Sharding an Injection Catalog for ApPipeWithFakes
+-------------------------------------------------
+
+When an injection catalog has already been ingested in unsharded form, use
+``scripts/fakes/shard_fake_catalogs.py`` to split and ingest sharded catalogs
+into an output collection before running ``ApPipeWithFakes``.
+
+The command supports Butler ``where`` expressions through ``-d/--dataquery``
+using the same syntax style as ``pipetask -d``.
+
+Example:
+
+.. code-block:: bash
+
+    python ${AP_PIPE_DIR}/scripts/fakes/shard_fake_catalogs.py \
+        -b /path/to/repo/butler.yaml \
+        -i u/user/input_collection \
+        -o u/user/output_collection \
+        -t VisitDetectorFakeSourceCat \
+        -d "instrument='LSSTComCam' AND skymap='lsst_cells_v2' AND visit=2024111100094 AND detector IN (0,3)"
+
+Arguments:
+
+* ``-b/--butler-config``: Butler config file path.
+* ``-i/--input-collection``: Collection containing unsharded injection catalogs.
+* ``-o/--output-collection``: Output collection for sharded catalogs.
+* ``-t/--dataset-type-name``: Dataset type to query (default: ``injection_catalog``).
+* ``-d/--dataquery``: Optional Butler ``where`` expression for dataset selection.
--- a/pipelines/LSSTComCam/ApPipeWithFakes.yaml
+++ b/pipelines/LSSTComCam/ApPipeWithFakes.yaml
@@ -19,3 +19,15 @@ parameters:
   coaddName: goodSeeing
   # APDB config file must be user-specified
   apdb_config:
+
+tasks:
+  isr:
+    class: lsst.ip.isr.IsrTaskLSST
+    config:
+      connections.outputExposure: post_isr_image
+      # Some instruments have this already in obs_lsst.
+      # Here we turn if off for AP generally.
+      doBrighterFatter: False
+      # Turn off the amp offset calculation.
+      # It is only used to compute metrics, and takes several seconds.
+      doAmpOffset: False
--- a/pipelines/LSSTComCam/ApPipeWithFakes.yaml
+++ b/pipelines/LSSTComCam/ApPipeWithFakes.yaml
@@ -26,7 +26,7 @@ tasks:
     config:
       connections.outputExposure: post_isr_image
       # Some instruments have this already in obs_lsst.
-      # Here we turn if off for AP generally.
+      # Here we turn it off for AP generally.
       doBrighterFatter: False
       # Turn off the amp offset calculation.
       # It is only used to compute metrics, and takes several seconds.
--- a/pipelines/_ingredients/injection/PostInjectedTasksApPipe.yaml
+++ b/pipelines/_ingredients/injection/PostInjectedTasksApPipe.yaml
@@ -1,11 +1,7 @@
-description: AP Pipeline with synthetic/fake sources. Templates are inputs.
-# This pipeline is imported by camera-specific pipelines.
-# You almost certainly want to run one of those, and not this one.
-
-# Per RFC-997, all variants of ApPipe must define three mutually exclusive subsets:
-# - preload is tasks that can be run before raw images arrive
-# - prompt is everything done by Prompt Processing starting from raws
-# - afterburner is metrics and other non-essential tasks that are skipped by Prompt Processing
+description: Auxiliar Pipeline file with synthetic/fake sources tasks.
+# This pipeline is used by source injection pipeline generation tools.
+# This is not a functional pipeline on its own, and is not intended to be run
+# directly.
 
 imports:
   - location: $ANALYSIS_TOOLS_DIR/pipelines/apDetectorVisitFakesCore.yaml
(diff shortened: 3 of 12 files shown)


python/lsst/ap/pipe/createApFakes.py
```diff
@@ -260,9 +286,461 @@ def createRandomMagnitudes(self, nFakes, rng):
...
+        Returns
+        -------
+        outputCat : `astropy.table.Table`
+            Catalog of fake sources to draw inputs from.
+        """
+        # Use the visit+detector ids as the random seed.
+        visitId = visit_image.getInfo().getVisitInfo().id
+        detId = visit_image.detector.getId()
+        seed = int(str(visitId)+str(detId))
+        rng = np.random.default_rng(seed)
```
isullivan:
```suggestion
        rng = np.random.default_rng([visitId, detId])
```
I am very wary of relying on string concatenation for creating the seed.

python/lsst/ap/pipe/createApFakes.py
```diff
@@ -260,9 +286,461 @@ def createRandomMagnitudes(self, nFakes, rng):
...
+    This task creates a catalog of fake sources that can be used to inject
+    sources into visit detector images.
+    """
+
+    _DefaultName = "createVisitDetectorFakes"
+    ConfigClass = CreateVisitDetectorFakesConfig
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.log = logging.getLogger(__name__)
```
isullivan, line 406: Not needed, this over-writes the inherited logger.

python/lsst/ap/pipe/createApFakes.py
```diff
@@ -241,8 +247,28 @@ def createVisitCoaddSubdivision(self, nFakes):
...
+            Dictionary containing per-band magnitudes plus a ``mag`` column
+            compatible with ``source_injection`` catalogs.
+        """
+        randMags = {}
+        for fil in self.config.filterSet:
+            randMags[self.config.mag_col % fil] = mags
+        randMags["mag"] = mags
+        return randMags
+
     def createRandomMagnitudes(self, nFakes, rng):
```
isullivan: It looks like the only use of `createRandomMagnitudes` was deleted earlier. Should this method be removed?

python/lsst/ap/pipe/createApFakes.py
```diff
@@ -260,9 +286,461 @@ def createRandomMagnitudes(self, nFakes, rng):
...
+                used_ids=catalog["injection_id"],
+            )
+            # create column of isVariable flag
+            catalog["isVariable"] = np.where(np.isin(np.arange(len(catalog)), idx), True, False)
+            variable_fakes["isVariable"] = True
+
+            catalog = vstack([catalog, variable_fakes])
+
+        if len(catalog) > len(np.unique(catalog["injection_id"])):
+            self.log.warning("Duplicate injection IDs detected after catalog assembly; reassigning them.")
```
isullivan, line 628: This might be worth raising an error if there are duplicates, instead of warning and reassigning.

python/lsst/ap/pipe/createApFakes.py
```diff
@@ -260,9 +286,461 @@ def createRandomMagnitudes(self, nFakes, rng):
...
+    def run(self, sourceCat, visit_image):
+        """Create a set of visit detector fakes.
+
+        Parameters
+        ----------
+        sourceCat : `lsst.afw.table.SourceCatalog`
+            Catalog of sources detected on the calibrated exposure.
+        visit_image : `lsst.afw.image.Exposure`
+            Visit image to inject synthetic sources into.
+        summaryMetrics : `lsst.metrics.MetricMeasurementBundle`
```
isullivan: `summaryMetrics` is not a parameter. Delete?

pipelines/_ingredients/injection/PostInjectedTasksApPipe.yaml
```diff
@@ -17,20 +13,20 @@ tasks:
   injectedMatchDiaSrc:
     class: lsst.pipe.tasks.matchDiffimSourceInjected.MatchInjectedToDiaSourceTask
     config:
-      connections.coaddName: parameters.coaddName
-      connections.matchDiaSources: "fakes_{coaddName}Diff_matchDiaSrc"
-      connections.injectedCat: fakes_preliminary_visit_image_catalog
+    - connections.coaddName: parameters.coaddName
```
isullivan: Drop the leading `-`

pipelines/_ingredients/injection/PostInjectedTasksApPipe.yaml
```diff
@@ -17,20 +13,20 @@ tasks:
...
+      connections.injectionCat: VisitDetectorFakeSourceCat
       connections.diffIm: difference_image
       connections.diaSources: dia_source_unfiltered
       matchDistanceArcseconds: 0.5
       trimBuffer: 50
   injectedMatchAssocDiaSrc:
     class: lsst.pipe.tasks.matchDiffimSourceInjected.MatchInjectedToAssocDiaSourceTask
     config:
-      connections.coaddName: parameters.coaddName
+    - connections.coaddName: parameters.coaddName
```
isullivan: Drop the leading `-`

doc/lsst.ap.pipe/ap-pipe-pipeline-with-fakes.rst
```diff
@@ -61,3 +61,39 @@ The key additional outputs include:
 * ``fakes_preliminary_visit_image_catalog``: Catalog of injected source properties
 * ``fakes_goodSeeingDiff_matchDiaSrc``: Matches between injected and detected sources
 * ``fakes_goodSeeingDiff_matchAssocDiaSrc``: Matches for successfully associated sources
+
+
+Creating an Injection Catalog
+-----------------------------
+To create an injection catalog for use with ``ApPipeWithFakes``, use the ``createInjectionCatalog`` command from the ``lsst.source.injection`` package. This command generates a catalog of synthetic sources based on specified parameters such as magnitude distribution, spatial distribution, and source types.
```
isullivan: Could you add an example command using `createInjectionCatalog` to create an injection catalog? A link to documentation with a working example would be fine, too.

doc/lsst.ap.pipe/ap-pipe-pipeline-with-fakes.rst
```diff
@@ -61,3 +61,39 @@ The key additional outputs include:
...
+Alternatively, you can use the ``createApFakes`` task from the ``lsst.ap.pipe`` package, which is specifically designed to create injection catalogs tailored for the AP pipeline.
```
isullivan, line 69: Also provide an example of using `createApFakes`.

doc/lsst.ap.pipe/ap-pipe-pipeline-with-fakes.rst
```diff
@@ -61,3 +61,39 @@ The key additional outputs include:
...
+-------------------------------------------------
+
+When an injection catalog has already been ingested in unsharded form, use
+``scripts/fakes/shard_fake_catalogs.py`` to split and ingest sharded catalogs
+into an output collection before running ``ApPipeWithFakes``.
+
+The command supports Butler ``where`` expressions through ``-d/--dataquery``
+using the same syntax style as ``pipetask -d``.
+
+Example:
```
isullivan, line 92: Thank you for the code example!

pipelines/LSSTComCam/ApPipeWithFakes.yaml
```diff
@@ -19,3 +19,15 @@ parameters:
...
   # APDB config file must be user-specified
   apdb_config:
+
+tasks:
+  isr:
+    class: lsst.ip.isr.IsrTaskLSST
+    config:
+      connections.outputExposure: post_isr_image
+      # Some instruments have this already in obs_lsst.
+      # Here we turn if off for AP generally.
```
isullivan: "if" -> "it"

python/lsst/ap/pipe/createApFakes.py
```diff
@@ -260,9 +286,461 @@ def createRandomMagnitudes(self, nFakes, rng):
...
+        pipelineConnections=CreateVisitDetectorFakesConnections
+):
+    """Config for CreateVisitDetectorFakesTask."""
+    randomFakeDensity = pexConfig.RangeField(
+        doc="Goal density of visit detector fake sources per square degree.",
+        dtype=float,
+        default=1000,
+        min=1,
+    )
+    nRandomFakes = pexConfig.RangeField(
```
isullivan, line 308: Very minor, but I don't think this needs to be a `RangeField`

python/lsst/ap/pipe/createApFakes.py
```diff
@@ -260,9 +286,461 @@ def createRandomMagnitudes(self, nFakes, rng):
...
+        doc="Whether to add model fakes to the visit detector.",
+        dtype=bool,
+        default=False,
+    )
+    magMin = pexConfig.RangeField(
+        doc="Minimum magnitude for the fake sources.",
+        dtype=float,
+        default=20,
+        min=16,
+        max=24,
```
isullivan: I could imagine someone wanting to inject fakes outside of these magnitude ranges. Probably drop the RangeField and min/max

python/lsst/ap/pipe/createApFakes.py
```diff
@@ -260,9 +286,461 @@ def createRandomMagnitudes(self, nFakes, rng):
...
+    )
+    magMax = pexConfig.RangeField(
```
isullivan: As above, probably drop the RangeField

python/lsst/ap/pipe/createApFakes.py
```diff
@@ -260,9 +286,461 @@ def createRandomMagnitudes(self, nFakes, rng):
...
+        # set of catalogs to concatenate at the end
+        catalog_set = []
+
+        photoCalib = visit_image.getPhotoCalib()
+        bbox = visit_image.getBBox()
+        xmin, xmax = bbox.getMinX(), bbox.getMaxX()
+        ymin, ymax = bbox.getMinY(), bbox.getMaxY()
+        visit_stats = visit_image.getInfo().getSummaryStats()
+
+        # corners = visit_image.getBBox().getCorners()
```
isullivan: Delete commented-out code?

python/lsst/ap/pipe/createApFakes.py
```diff
@@ -260,9 +286,461 @@ def createRandomMagnitudes(self, nFakes, rng):
...
+            n_variable_fakes = int(len(catalog) * self.config.variableFakeFraction)
+            idx = rng.choice(len(catalog), size=n_variable_fakes, replace=False)
+            variable_fakes = catalog[idx].copy()
+            variable_fakes["mag_offset"] = rng.normal(
+                loc=self.config.variableFakeMean,
+                scale=self.config.variableFakeStd,
+                size=n_variable_fakes
+            )
+            variable_fakes["mag"] += variable_fakes["mag_offset"]
+            variable_fakes["isVisitSource"] = ~variable_fakes["isVisitSource"]
```
isullivan, line 613: Worth a code comment explaining the flag negations here.

python/lsst/ap/pipe/createApFakes.py
```diff
@@ -260,9 +286,461 @@ def createRandomMagnitudes(self, nFakes, rng):
...
+                    [id_map.get(int(twin_id), int(twin_id)) for twin_id in catalog["twin_id"]],
+                    dtype=np.int64,
+                )
+
+        catalog["visit"] = visitId
+        catalog["detector"] = detId
+
+        return Struct(outputCat=catalog)
+
+    def select_hosts(self, sourceCat, *args, **kwargs):
```
isullivan: Why include `*args` and `**kwargs` if they're not used?

isullivan (approved):
Looks good! A few comments for readability, and potential bugs.

python/lsst/ap/pipe/createApFakes.py
BrunoSanchez: But we should not allow people to inject values that are none sense right? I think people will want to have a range of magnitudes. I don't know if it is wise to drop the field completely. Maybe increase the range?

isullivan: I suppose, but there isn't a physical hard limit that makes sense. If we set limits here, then it is not possible for someone using our pipelines to process data from a different observatory to overwrite the magnitude ranges in the config to anything brighter or fainter.

BrunoSanchez: What I don't want to is to leave this as non-configurable. How do we do that? Set instead of a RangeField a free value parameter field? Not sure if that's your suggestion here.

isullivan: I was just thinking of making it a standard `Field`, since it has a reasonable default value.

python/lsst/ap/pipe/createApFakes.py
BrunoSanchez: I disagree, we are creating the injection ids here, so we just need to fix this. NO need to raise an error I think.

PR #276, opened 2026-06-06
DM-55171: Add consolidateDiaSourceVisit to ApPipe

Gerenjie:


pipelines/_ingredients/ApPipe.yaml
```diff
@@ -66,6 +66,11 @@ tasks:
       connections.calexp: preliminary_visit_image
       connections.visitSummary: preliminary_visit_summary
       connections.visitSummarySchema: preliminary_visit_summary_schema
+  consolidateDiaSourceTable:
```
isullivan: This task has the output of `standardizeDiaSource` as its input, so it needs to be moved after that task in the definitions. Also, add a contract to check that the `connections.inputCatalogs` input connection matches the `connections.diaSourceTable` output of `standardizeDiaSource`

isullivan (approved):
See the comment on re-ordering the new task in ApPipe.yaml, and adding a new contract for the connections.

Also note that `consolidateDiaSourceTable` needs to be removed from `Conversions.yaml` and `ConversionsForFakes.yaml` from `ap_verify`

Gerenjie:
Is this good to merge?

isullivan:
I marked the PR as `approved` when I left my review, so it is OK to merge when you are satisfied you have addressed my comments. I don't need to look at it again.

PR #278, opened 2026-06-25
DM-55233: Add standardization steps for prompt_source

isullivan:
Includes one unrelated change that redefines how the isr task is set for LSSTCam pipelines. It was necessary to update the pipeline definition so that the new subset definitions could be inherited properly, without needing to duplicate changes.


jrmullaney (approved):
LGTM

PR #280, opened 2026-06-30
DM-55341: Optimizations for daily batch

ctslater:
* Create bigger clusters, instead of per-visit-detector quanta.
* Prune `getRegionTimeFromVisit` in addition to `loadDiaCatalogs`.

### DIFF
--- a/scripts/LSSTCam/submit_ap_daytime.sh
+++ b/scripts/LSSTCam/submit_ap_daytime.sh
@@ -104,9 +104,10 @@ DATA_QUERY="instrument='$INSTRUMENT' \
         --qgraph-datastore-records \
         -q "$FULL_QGRAPH"
 
-    echo "[$(date)] Step 2/3: pruning orphan loadDiaCatalogs quanta"
+    echo "[$(date)] Step 2/3: pruning orphan quanta"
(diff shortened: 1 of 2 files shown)


scripts/LSSTCam/submit_ap_daytime.sh
```diff
@@ -103,9 +103,10 @@ DATA_QUERY="instrument='$INSTRUMENT' \
         --dataset-query-constraint off \
         -q "$FULL_QGRAPH"
 
-    echo "[$(date)] Step 2/3: pruning orphan loadDiaCatalogs quanta"
+    echo "[$(date)] Step 2/3: pruning orphan quanta"
     python -m lsst.ap.pipe.prune_orphan_preloads \
-        "$FULL_QGRAPH" "$PRUNED_QGRAPH"
+        "$FULL_QGRAPH" "$PRUNED_QGRAPH" \
+        --preload-labels getRegionTimeFromVisit --anchor-label associateApdb
```
tcjennings: The argument is `--preload-label`, no plural.

ctslater: Thanks, I had pluralized it when I added support for multiple labels, but forgot to revert when I dropped that change.

PR #281, opened 2026-07-03
DM-55369: Add singleFrameDetectAndMeasure to SingleFrame

Gerenjie:

PR #283, opened 2026-07-08
Revert "Merge pull request #281 from lsst/tickets/DM-55369"

ebellm:
`singleFrameDetectAndMeasure` should only run in daytime AP, which it does via the `afterburner` subset of ApPipe.yaml


hsinfang (approved):

hsinfang:
Currently Daytime AP uses the yaml directly: 
https://github.com/lsst/ap_pipe/blob/w.2026.27/bps/LSSTCam/bps_Daytime.yaml#L1 

not the `afterburner` subset.  But doesn't matter here.

PR #286, opened 2026-08-05
DM-55729: Fix bitrot in PromptTemplate

erinleighh:


mrawls (approved):
Agree the tasks had changed and should be correctly represented now. Thanks for doing a BPS mini-run to ensure it works. It's too bad this can't be tested directly in test_pipeline.py due to dependencies on drp_pipe, but we don't have to fix that today.

PR #287, opened 2026-08-06
DM-55738: Update PromptTemplate to inherit from QuickTemplate

erinleighh:


pipelines/_ingredients/QuickTemplate.yaml
```diff
@@ -25,10 +33,11 @@ tasks:
       # take many 10s of seconds in crowded fields
       do_remeasure_star_background: False
       doMaskDiffractionSpikes: True
-  selectDeepCoaddVisits:
+  selectTemplateCoaddVisits:
```
mrawls: This already exists exactly how we want it to be (as a default, anyway) in the DRP pipeline, so you can just import it here without any config overrides.

pipelines/_ingredients/QuickTemplate.yaml
```diff
@@ -37,23 +46,37 @@ tasks:
   assembleTemplateCoadd:
     class: lsst.drp.tasks.assemble_coadd.CompareWarpAssembleCoaddTask
     config:
-      connections.selectedVisits: deep_coadd_visit_selection
+      connections.selectedVisits: template_coadd_visit_selection
```
mrawls: This is default and does not need to be specified here

pipelines/_ingredients/QuickTemplate.yaml
```diff
@@ -6,10 +6,18 @@ imports:
...
   - consolidateVisitSummary
-  - selectDeepCoaddVisits
+  - selectTemplateCoaddVisits
   - makeDirectWarp
   - makePsfMatchedWarp
   - assembleTemplateCoadd
+  - makeBinnedTemplateCoaddImage
+  - makeBinnedTemplateCoaddNImage
+  - makeWholeTractTemplateCoaddImage
+  - makeWholeTractTemplateCoaddNImage
```
mrawls, line 18: Please add the following (here and below in the subsets as appropriate):
- aggregateTemplateCoaddMaskFractions
- makeAggregatedTemplateCoaddMaskFractionsTable
- aggregatedTemplateCoaddMaskFractionsWholeSkyPlot
- makeTemplateCoaddInputSummaryTract
- makeTemplateCoaddInputSummary

pipelines/LSSTCam/PromptTemplate.yaml
```diff
@@ -4,57 +4,19 @@ description: The LSSTCam AP template building pipeline with preliminary_visit_im
...
+- location: $AP_PIPE_DIR/pipelines/_ingredients/QuickTemplate.yaml
+  exclude:
+  - isr
+  - calibrateImage
 
 tasks:
   consolidateVisitSummary:
     class: lsst.pipe.tasks.postprocess.ConsolidateVisitSummaryTask
     config:
       full: True
```
mrawls, line 17: Consider adding a comment for why this override is needed (because we're starting with preliminary_visit_images?)

pipelines/LSSTCam/PromptTemplate.yaml
```diff
@@ -4,57 +4,19 @@ description: The LSSTCam AP template building pipeline with preliminary_visit_im
...
-  makeDirectWarp:
-    class: lsst.drp.tasks.make_direct_warp.MakeDirectWarpTask
-    config:
-      connections.visit_summary: preliminary_visit_summary
-      useVisitSummaryPsf: False
   selectTemplateCoaddVisits:
     class: lsst.pipe.tasks.selectImages.BestSeeingSelectVisitsTask
```
mrawls, line 21: Add a comment that you are explicitly using a different class here on purpose

mrawls: Or use a different label, e.g., `selectTemplateBestSeeingSelectCoaddVisits`?

mrawls (commented):
Once these comments are addressed I think the pipelines will have the desired configurations - remains to be seen if they actually create templates we like 😄

